PR #69, opened 2012-06-10
Do not escape contents within 'style' tags.

favonia:
CSS selectors use `>` and `"`, but they get escaped in the default TagSoup renderer. See [here](http://www.w3.org/TR/CSS2/selector.html#pattern-matching) for CSS2 selector syntax.

### DIFF
--- a/src/Hakyll/Web/Urls.hs
+++ b/src/Hakyll/Web/Urls.hs
@@ -8,6 +8,7 @@ module Hakyll.Web.Urls
     ) where
 
 import Data.List (isPrefixOf)
+import Data.Char (toLower)
 import System.FilePath (splitPath, takeDirectory, joinPath)
 import qualified Data.Set as S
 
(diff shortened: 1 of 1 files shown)


gwern:
Has there been a regression on this? I see no test for quotes or greater-than, and I can't get Hakyll to not break inline `<style>` tags using these. I get HTML escaping of both of them: `<style>@media only screen and (max-width:64.9ch){#TOC a{display:inline-block}#TOC&gt;ul&gt;li&gt;ul{column-count:1}</style>`.

gwern:
On further investigation (ironically, I was actually looking at how to add the new `loading="lazy"` Chrome feature for native lazy loading of images, which works very nicely and is something Hakyll might want to consider as a default, since it is backwards compatible and saves a *ton* of bandwidth on something like [my StyleGAN page](https://www.gwern.net/Faces)), I've realized this problem was not Hakyll at all but was providing a hint: it was actually *Tagsoup*, and as the function notes, Tagsoup by default screws up stuff in the `<style>` tags by escaping them. Oops. Never mind...

Issue #152, opened 2013-05-15
Allow a global metadata file


jaspervdj:
Some people prefer putting all of the important metadata in a single file instead of splitting it out. It would be nice if we could either have a "global" or per-directory metadata file. The format should be close to the current format, so probably something like:

    --- foo/bar.markdown
    title: bar
    author: Jasper

    --- foo/qux.rst
    title: jiggles
    author: Jimmy

rasendubi:
@jaspervdj see https://github.com/jaspervdj/hakyll/issues/152#issuecomment-60727025

PR #190, opened 2013-10-12
Implemented per-directory metadata support

krsch:
Implemented per-directory metadata file as it was mentioned in #152.
Format is closed to the one described in the issue, but after `---` I have a glob pattern terminated by newline (space at the end are considered a part of the pattern). The path in this pattern is relative to the directory with the metadata file.
Metadata in old locations (embedded and *.metadata) override per-directory metadata and the one in a subdirectory overrides the one in a parent directory.


jtojnar:
Why was this reverted?

PR #300, opened 2014-09-17
matchMetadata for examining metadata when defining rules

ohbadiah:
Someone else on the mailing list asked last month about using metadata to decide whether or not to compile something when defining `Rules`. This capability would solve a problem for me, too.

I realized all of the tools were already there in the form of the `getAllMetadata` function on the `MonadMetadata` typeclass. This PR just creates an alternative to `match` that also takes a predicate on metadata.

### DIFF
--- a/src/Hakyll/Core/Rules.hs
+++ b/src/Hakyll/Core/Rules.hs
@@ -19,6 +19,7 @@
 module Hakyll.Core.Rules
     ( Rules
     , match
+    , matchMetadata
     , create
     , version
     , compile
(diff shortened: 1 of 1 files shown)


jaspervdj:
Awesome, thanks!

eunchan:
So, with this commit, can we opt out using metadata? For example, if 'public' is not set 'true', can matchMetadata function skip the compile?

It would be really helpful if you give us some examples with this function. Thanks for updating!

ohbadiah:
What you're talking about should definitely be possible, with something like:

```haskell
import Hakyll
import qualified Data.Map as Map 

metadataFieldIs :: String -> String -> Metadata -> Bool
metadataFieldIs key value metadata = case Map.lookup key metadata of
  Just v  -> value == v
  Nothing -> False

postIsPublic :: Metadata -> Bool
postIsPublic = metadataFieldIs "public" "true"

matchPublicPosts = matchMetadata "content/posts/*" postIsPublic
```

My own use case is just a little more complicated. When compiling posts for the "food" subblog, I don't want to pick up posts that don't have "food" as one of the words in the "subblog" `Metadata` field. You can see the code where I'm doing this [here](https://github.com/ohbadiah/nickmcavoy-dotcom/blob/master/blog/src/Main.hs#L104), with a helper I define [here](https://github.com/ohbadiah/nickmcavoy-dotcom/blob/master/blog/src/Nick/Subblogs.hs#L23).

eunchan:
@ohbadiah Thanks for the example. I am eager to get the new release!

jaspervdj:
I'll try to do a release today!

jaspervdj:
Released on Hackage as `hakyll-4.6.4.0`.

PR #310, opened 2014-11-02
Bump pandoc-citeproc upper bound

bgamari:


bgamari:
Looks like this was done in another patch.

PR #330, opened 2015-02-11
Remove test debug output

paulkeene:
When running hakyll's test suit locally I noticed that a file named "foo" is left in the parent directory of the project after the test suit completes. It looks like this was added for debugging purposes at some point and was accidentally committed.


jaspervdj:
Good catch, thanks!

PR #335, opened 2015-02-24
Don't add .cabal extension twice in init

rootzlevel:
Previously, hakyll-init would create the cabal file with a name like
name.cabal.cabal.


jaspervdj:
Oh, great, thanks for fixing!

PR #339, opened 2015-03-23
let pandoc handle DocBook files

kadoban:
This is pretty close to completely untested, but may work. I'm submitting it as a pull request only because it may be useful to the orignal asker of this question: https://groups.google.com/forum/#!topic/hakyll/Y2oSt2smrn8


jaspervdj:
Thanks!

PR #342, opened 2015-03-30
Added support for a custom teaser separator

NorfairKing:
This way you don't need to use the  separator.
You can define your own.


jaspervdj:
Thanks @NorfairKing for the PR and @lubomir for noticing the typo!

PR #349, opened 2015-05-19
Keep .git when cleaning destinationDirectory

netzwerg:
When publishing a Hakyll-based site to GitHub, it is very convenient to clone the `master`/`gh-pages` branch of the destination repo directly to the `_site` folder.

Taking my own blog as an example:
```
git clone -b source https://github.com/netzwerg/netzwerg.github.com.git 
cd netzwerg.github.com
git clone -b master https://github.com/netzwerg/netzwerg.github.com.git _site 
``` 
This allows to edit the sources, build the site, and publish the final pages from a single directory tree.

Unfortunately, `cabal run clean` wipes the complete `_site` folder (including the `.git` folder required for the above strategy to work).

The suggested change keeps a potentially existing `_site/.git` folder upon `clean`.

*Disclaimer:* I am new to Hakyll/Haskell, please refactor/refine/rewrite as needed -- tia!

### DIFF
--- a/src/Hakyll/Commands.hs
+++ b/src/Hakyll/Commands.hs
@@ -16,9 +16,11 @@ module Hakyll.Commands
 --------------------------------------------------------------------------------
 import           Control.Applicative
 import           Control.Concurrent
-import           Control.Monad              (void)
+import           Control.Monad              (void, forM_)
 import           System.Exit                (ExitCode, exitWith)
 import           System.IO.Error            (catchIOError)
+import           System.Directory           (doesDirectoryExist, getDirectoryContents, removeFile)
+import           System.FilePath            ((</>), pathSeparator)
 
 --------------------------------------------------------------------------------
 import qualified Hakyll.Check               as Check
@@ -56,7 +58,7 @@ check config logger check' = Check.check config logger check' >>= exitWith
 -- | Remove the output directories
 clean :: Configuration -> Logger -> IO ()
 clean conf logger = do
-    remove $ destinationDirectory conf
+    removeGitFriendly $ destinationDirectory conf
     remove $ storeDirectory conf
     remove $ tmpDirectory conf
   where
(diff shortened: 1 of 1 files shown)


nagisa:
Such change was proposed multiple times and it was declined multiple times as well on the grounds of 

* `destinationDirectory` being owned by hakyll;
* There existing multiple ways to track files inside directory without putting `.git` into it;
* many more special directories that might want to not be deleted.

netzwerg:
Thanks for your clarifications. I must have missed the issues that already dealt with similar suggestions (or is there a mailing-list where these discussions are happening?).

nagisa:
I believe none of the past proposals have moved past message on IRC (and we, sadly, keep no logs). 

Keep in mind, that I have no decision-making power in this project and am just stating opinion that was prevalent a few months ago; I’d reopen this for now and see what @jaspervdj has to say on this 😊

netzwerg:
Ok, l have re-opened it -- let's wait and see :-)

jaspervdj:
I think you can use `$GIT_DIR` and `$GIT_WORK_DIR` for this, but I'm open to discussion. Thanks for handling this so far @nagisa.

cryptica:
Keeping the .git folder in the destination directory would also allow to keep it as a submodule in the gh-pages branch of the same repo (as in https://github.com/shiftyp/submodule-gh-pages-example), therefore separating source and built files while keeping easy deployment to GitHub. I don't think that works easily with $GIT_DIR and $GIT_WORK_DIR.

As for other special directories, maybe all . directories within the destination directory could be kept?

listx:
FWIW, I use rsync as part of a post-build script, like this:

```
dest='listx.github.io/'
targ='_site/'
rsync -ahP --no-whole-file --inplace --delete --exclude=.git $targ $dest                                                                                                           
cp -f .gitignore $dest
```

where `listx.github.io` is its own git repo.

I think if we do go ahead with keeping the `.git` folder (or any other special folders) within `_site` untouched, we should make the user define all such special cases explicitly; i.e., let's not implicitly treat some particular files/paths as special.

cryptica:
I like the idea of keeping the built files in the same repository, but on a separate branch. By including it as a submodule, there is need to call cp or rsync in a post-build script.

Adding an option to specify which directories to keep within `_site` would also work and be perfectly acceptable for me.

PR #351, opened 2015-05-27
Add getItemModificationUTC

jaiyalas:
This function allows to obtain the modification time in UTC which can be useful in introducing a modification-time-related customized field. Personally I need this to convert UTCTime into ZonedTime or LocalTime.


jaspervdj:
This is already in there as `getItemModificationTime`.

PR #352, opened 2015-06-14
add support for mediawiki

aleiphoenix:
I choose Pandoc for my doc parser because it supports mediawiki. I'm loving Haskell and prefer writing my posts in *.mediawiki. So I add this support in Hakyll :)


jaspervdj:
Thanks!

PR #368, opened 2015-08-16
Add "poster" as a URL attribute

vtduncan:
Similar to #134, this change enables the `poster` attribute of videos to be recognized as a URL and processed using `withURLs`. Here's where it's referenced in the [HTML standard](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-video-poster).

Thanks for maintaining hakyll!


jaspervdj:
Thanks!

PR #374, opened 2015-09-24
prevent hakyll-init from generating directory with leading hyphen(s)

Javran:
When running `hakyll-init --help`, the current behavior is to generate a directory of `--help`.
But I think a more intended behavior would be providing help messages rather than creating the directory.


jaspervdj:
Great! My only reservation is that the `isPrefixOf` is not completely self-explanatory. Would you mind adding a comment there?

Javran:
comment added, please check if it looks good :)

jaspervdj:
Thanks!

PR #379, opened 2015-10-12
sigkill.dk now uses Hakyll 4.

Athas:
I also updated [my tutorial](http://sigkill.dk/writings/guides/hakyll.html), but I'm not really sure where it would fit.


jaspervdj:
Thanks!

PR #385, opened 2015-10-21
templateCompiler compiles resource body only

### LINKED ISSUE #378: templateCompiler should only parse the item body
I want (need) to use metadata is (some of) my templates.
However, the default `templateCompiler` honors this by including the metadata in the `Template` (which renders it into the page), I had to write my own one - though that was fairly easy at least :-)

It should be as simple as replacing
```hs
    item <- getResourceString
```
with
```hs
    item <- getResourceBody
```

bergus:
fixes #378

### DIFF
--- a/data/example/site.hs
+++ b/data/example/site.hs
@@ -57,7 +57,7 @@ main = hakyll $ do
                 >>= loadAndApplyTemplate "templates/default.html" indexCtx
                 >>= relativizeUrls
 
-    match "templates/*" $ compile templateCompiler
+    match "templates/*" $ compile templateBodyCompiler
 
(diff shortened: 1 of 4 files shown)


jaspervdj:
Looks good, but could we do this in an additional `templateBodyCompiler` function? Some people *might want* the metadata there, depends on what sort of files they're generating.

bergus:
No problem. I don't think there's (m)any of them, but we don't want to make this a breaking change of course.
Commit is updated.

jaspervdj:
Forgot about this somehow. Thanks for taking the time to fix this.

PR #389, opened 2015-11-20
Item context has precedence before feed context

cblp:
The item's author is the item's author, not the feed's owner.


jaspervdj:
Thanks!

PR #402, opened 2016-02-06
Add Snippet expression for snippet includes

Nicowcow:
I've been using a `$snippet("...")$` for some time, which I find quite useful when I don't want my included files to be processed like templates, as happens with `$partial("...")$`. Typical use case, I want to include an extra Haskell file which contains `$`s and `<$>`s  and most of the file is eaten away. Might be useful to other people as well.

It requires something like this in the rules:

``` haskell
      match "snippet.txt" $ compile getResourceBody
```

instead of

``` haskell
      match "template.txt" $ compile templateCompiler
```

Let me know if I missed an existing solution; also, no hard feelings if this gets rejected. If there's a chance this gets in, I'll update with the doc in `Template/Internal`.

### DIFF
--- a/src/Hakyll/Web/Template/Context.hs
+++ b/src/Hakyll/Web/Template/Context.hs
@@ -18,6 +18,7 @@ module Hakyll.Web.Template.Context
     , urlField
     , pathField
     , titleField
+    , snippetField
     , dateField
     , dateFieldWith
     , getItemUTC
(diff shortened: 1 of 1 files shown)


jaspervdj:
It's not necessary to implement this by patching the `Template` internals. You can use `functionField` instead. In your `site.hs`:

```haskell
myCtx :: Context String
myCtx =
    functionField "snippet" snippet <>
    defaultContext
  where
    snippet [thingToInclude] _ = loadBody (fromFilePath thingToInclude)
```

And then use `myCtx` wherever you want `$snippet(..)$` to be available.

This looks like a useful function though, I wouldn't mind a pull request which adds this function as `snippetField` to `Hakyll.Web.Template.Context`.

Nicowcow:
Way cleaner, thanks! I wish I had seen `functionField` before. I'll have a look at it tonight and modify the pull request some time this week.

Nicowcow:
Updated, though not sure why travis failed

jaspervdj:
Thanks!

PR #406, opened 2016-02-29
Renew example posts' dates

raindev:
I think it makes sense to use newer dates in the generated example posts to not leave impression of obsolescence.


jaspervdj:
Right, makes sense to me.

PR #407, opened 2016-03-14
Make Travis badge Retina ready

raindev:
Use http://shields.io for a higher quality badge.

Here is a screenshot from my MacBook Pro highlighting the difference (new badge on the right):
screen shot 2016-03-14 at 08 28 03


jaspervdj:
Thanks!

PR #411, opened 2016-03-16
Recommend Stack in installation/basics tutorials

hdgarrood:
See also the mailing list thread: https://groups.google.com/forum/#!topic/hakyll/CR-k5YFi9WA

The only potential downside I can see to doing it this way is that you have to replace `./site <command>` with `stack exec site <command>`, and so this might 'break' other tutorials or materials that refer to Hakyll.


jaspervdj:
Looks good to me. I think we can assume that Hakyll/Haskell users are smart enough to figure out what stack is.

PR #418, opened 2016-04-25
Adds itemSetIdentifier to Item.hs and changeIdentifier to Compiler.hs.

lettier:
Hello @jaspervdj and contributors. :)

This change allows one to change/set the identifier of an Item.

Use case:

Say someone wants to show their most recent post on the home page in its entirety.

```haskell
-- /site.hs
-- ...
  create ["index.html"] $ do
    route idRoute
    compile $ do
      post <- fmap head . recentFirst =<< loadAllSnapshots "posts/*" "content"
      loadAndApplyTemplate "templates/default.html" postCtx post
        >>= relativizeUrls
        >>= changeIdentifier "index.html"
-- ...
```
```html
<!-- "templates/default.html" -->
<html>
  <head>
    <meta property="og:url" content="$url$">
    <title>$title$</title>
  </head>
  <body>
    $date$
    $body
  </body>
</html>
```

This will allow them to create or match a page with a different identifier than the one they wish to load. In this example they are loading say Identifier `posts/yyyy-mm-dd-my-recent-post.markdown` but they output the Identifier `index.html`. 

You can see it in use here https://lettier.github.io rendered and the [source code](https://github.com/lettier/lettier.github.io/blob/hakyll/site.hs#L51). 

Thanks!

### DIFF
--- a/src/Hakyll/Core/Compiler.hs
+++ b/src/Hakyll/Core/Compiler.hs
@@ -6,6 +6,7 @@ module Hakyll.Core.Compiler
     , getUnderlying
     , getUnderlyingExtension
     , makeItem
+    , changeIdentifier
     , getRoute
     , getResourceBody
     , getResourceString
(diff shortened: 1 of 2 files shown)


jaspervdj:
Hey, I don't think this is a good idea. The identifier should be immutable as much as possible. you can do something like this:

```
item <- loadAndApplyTemplate "templates/default.html" postCtx post >>= relativizeUrls
makeItem (itemBody item)
```

lettier:
Okay, that works too. Thanks @jaspervdj. :+1:

PR #432, opened 2016-06-02
Move to stack for travis CI testing.

sakshamsharma:
Slightly hacky travis script.
Here's a log from my test: [here](https://travis-ci.org/sakshamsharma/hakyll/builds/134875314)
Caching is enabled, so it should not build all the dependencies every time.
Also, you need to override some of the travis phases or else it will try to run cabal, which will break during dependency resolution (and is not going to be of any use anyway).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,20 @@
 language: haskell
 ghc: '7.10'
+branches:
+  only:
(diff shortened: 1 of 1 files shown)


jaspervdj:
Thanks!

PR #437, opened 2016-06-14
Allow data-default 0.7, pandoc-citeproc 0.10, and tagsoup 0.14

felixonmars:
Building with the new releases went fine here, and it works correctly.


jaspervdj:
Thanks!

PR #439, opened 2016-06-17
Loosen version restriction on data-default

toothbrush:
* hakyll.cabal (data-default): Allow data-default<0.8, since otherwise
  there's a version conflict when building with GHC8.0.1 and
  Cabal-1.24.0.0.


jaspervdj:
Thanks!

Issue #334, opened 2015-02-24
./site check is too verbose


gdetrez:
Right now, `./site check` prints a line for every checked page, even if there's no error. With a lot of pages, it becomes a bit painful to find the errors. I'd suggest to only print pages that have errors, unless maybe some debug mode is enabled.

jtojnar:
Some servers also return the whole response body in `X-Response-Body-Start` header. For example https://www.suomenlatu.fi/kiilopaa/en/accommodation/holiday-apartments/:

```
[...]

PR #440, opened 2016-06-25
Make errors in Check less verbose

jtojnar:
As reported in #334, sometimes the `Exception`s litter the terminal with unnecessary details. This commit will make errors output in the form `404 Not Found` for `StatusCodeException` and `SomeOtherException` for other exceptions.

I tried to honour the 80 character-limit; if you find some formatting or other issues or have a better idea how to get the exception name, I can rebase it.

Also it fixes `SIGINT`, thanks to @​Yuras on [StackOverflow](http://stackoverflow.com/a/38032304/160386).


jaspervdj-luminal:
Looks great, thanks!

PR #441, opened 2016-06-28
Fix rsync command in GH Pages tutorial

Pratyush:
This fixes the incorrect `rsync` command, as noted in [this Google Group post](https://groups.google.com/forum/#!topic/hakyll/PBtPQ0PK-20). The new command is more conservative and preserves the `.git` folder and the `.gitignore` file.


jaspervdj:
Thanks!

Issue #444, opened 2016-07-12
Port data/example/templates/default.html to HTML5


ethagnawl:
Would anyone be opposed to porting the templates/default.html file (generated by `hakyll-init`) to valid HTML5? 

I'd wager that a significant portion of the people using `hakyll-init` are doing that themselves, anyways. The folks who aren't (perhaps because they don't know/care about the difference) should _probably_ be using it, too. So, IMHO it'd be a win for everybody.

jaspervdj:
I think that's a very good idea.  Would you be up for doing this?

ethagnawl:
Sure thing.

ethagnawl:
Addressed in #445.

PR #445, opened 2016-07-12
#444 - porting default.html from XHTML to HTML5

ethagnawl:
- ported the default.html from XHTML to HTML5
- used [HTML5 Boilerplate](https://html5boilerplate.com/) as a starting point
- removed some of the additional niceties provided by Boilerplate like the favicon, touch icon, description meta tag, etc. but it might also make sense to introduce those (and the associated assets)
- it also might make sense to use more semantic markup (`<header>`, `<footer>`, etc.) in default.html

@jaspervdj How would I go about testing this locally? This will be my first contribution to a Haskell project. :)


jaspervdj:
This looks good to me.  If you want to test this locally, the easiest way is just to install it from the repo using `stack install` and then run `hakyll-init`.  That being said, I'm happy to merge this as it is now, and test it quickly myself.

data/example/templates/default.html
```diff
@@ -1,11 +1,11 @@
...
-<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+<!doctype html>
+<html lang="en">
     <head>
-        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+        <meta charset="utf-8">
+        <meta http-equiv="x-ua-compatible" content="ie=edge">
         <title>My Hakyll Blog - $title$</title>
-        <link rel="stylesheet" type="text/css" href="/css/default.css" />
+        <meta name="viewport" content="width=device-width, initial-scale=1">
```
crodjer: This meta tag could be moved above title, grouped with all other meta tags.

ethagnawl: @crodjer Good call.

ethagnawl:
@jaspervdj I was able to test it myself using the binary in .stack-work/install/... and everything looks good to me.

Here's the output: [default.html](https://gist.github.com/ethagnawl/7091bf7c3a9ac20c718c718aa0c1595c)

One small nitpick I noticed while running the output through a validator is that the img in index.html is missing an alt tag. I'd be happy to introduce it in this PR, create a separate issue or just ignore it.

jaspervdj:
Thanks again! :-)  Feel free to fix the alt tag in other PR if you want to do that.  I don't think it's super important though.

PR #447, opened 2016-07-22
Switch preview server from Snap to Warp

Arguggi:


Arguggi:
Forgot to run tests, will fix them asap

PR #453, opened 2016-07-24
cmdArgs to Options.Applicative

sk3r:
changed shell options parser from cmdAgrs to Options.Applicative


jaspervdj:
Thanks again for doing this!

Issue #414, opened 2016-04-17
$for template leaves trailing spaces behind


raindev:
This snipped

```
    $for(posts)$
        <a href="$url$">$title$</a> - $date$
    $endfor$
```

will result in this HTML being generated (`.` denotes trailing spaces)

```
....
        <a href="$url$">$title$</a> - $date$
....
```

This snipped will not result in any trailing spaces to the contrary

```
$for(posts)$
    <a href="$url$">$title$</a> - $date$
$endfor$
```

Use of the template should not be affected by the presence of spaces as they are useful for structuring the template code but not needed in the resulting HTML. Even the whole line with the template could be cut out while generating the result.

raindev:
The same applies to `if` template.

jaspervdj:
It's tricky to decide what to do here. The templates are not only used for rendering HTML (e.g., I also use them render markdown for some sites), which _is_ whitespace-sensitive.

I think one solution would be to have something like in Ruby's ERB, e.g.:

- `$foo$`: Renders the key `foo`
- `$-foo$`: Renders the key `foo` and removes leading whitespace from that line
- `$foo-$`: Avoids the line break after the expression

That way, the user can easily control this.

raindev:
Thanks for the timely reply, Jasper. You're right: current behavior isn't a bug exactly. The feature you have described seems like a reasonable design to implement. The issue could be closed, perhaps.

jaspervdj:
I'll leave it open as reminder to myself (or others) to implement this.

smorele:
I've found some cases with ending "-" like [here](https://github.com/jaspervdj/hakyll/blob/32e34f435c7911f36acdf4a62eec1f56faf0b269/tests/data/strip.html#L22)
Is this functionality implemented ?

bergus:
Looks like this can be closed now.

jaspervdj:
Thanks for the ping @bergus!

@smorele Yes, this functionality is now implemented.

PR #454, opened 2016-07-25
Whitespace trimming

samgd:
#414


samgd:
Will fix the conflic(s) but just about to fly home and then am back at work thus PR'd incase anyone wants it urgently :-)

src/Hakyll/Web/Template.hs
```diff
@@ -115,7 +115,31 @@
...
+--
+-- > listField "counts" (field "count" (return . itemBody))
+-- >    (sequence [makeItem "3", makeItem "2", makeItem "1"])
+--
+-- and a template
+--
+-- > <p>
+-- >     $for(counts)-$
+-- >       $count$
+-- >       $-sep-$...
```
samgd: This '-' isn't actually doing anything

samgd:
Worth noting for versioning that variables can no-longer have a trailing '-' if it's followed by a '$' as the '-' now "binds" to the '$' to signify stripping whitespace.

jaspervdj:
This looks great, thanks again!

src/Hakyll/Web/Template/Internal.hs
```diff
@@ -1,13 +1,12 @@
 --------------------------------------------------------------------------------
 -- | Module containing the template data structure
-{-# LANGUAGE DeriveDataTypeable         #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 module Hakyll.Web.Template.Internal
-    ( Template (..)
```
bergus: Hi, I just noticed these commits by chance.
Not exposing the `Template` constructor any more does break my code :-/ I have to admit that #384 was quite a hacky hack, and "Internal" *does* mean *internal*, but I'd love to keep inspecting templates via `unTemplate`.  Can you re-enable this, please?
Moving the `Template` data type to the base module of course makes it impossible to export it from the `Internal` module. Why was this necessary?

jaspervdj: The move was necessary because otherwise @samgd were running into cyclical module issues.  But if you need it we can expose the constructor from `Hakyll.Web.Template` or do some moving around such that the constructor is still exposed.

jaspervdj: @bergus I have rectified this here: 82ba9542e75238f4b69d1a497d429962cdff1e14

bergus: Ah, thanks a lot, that does look much cleaner!

PR #458, opened 2016-08-02
Enable `dateField` to put a timestamp with a local time zone

mnacamura:
I noticed that `dateField` enforces putting timestamps with UTC time. However, one may want to use a local time offset; at least I want to do so ;)
This PR achieves it.


jaspervdj:
This looks more or less good.  Suggestions:

1. I would put the `TimeLocale` as `Maybe TimeLocale` in `FeedConfiguration`.  Then, I would also add a `defaultFeedConfiguration` which sets this to `Nothing` (yes, I'm aware that this is a breaking change but it should be easy to fix for people).
2. Remove the `DateFormat` from the `dateField` function and just use the default one there.
3. Add a record which bundles up the settings for `dateFieldWith`:

        data DateFieldConfiguration = DateFieldConfiguration
            { dfcOutputTimeLocale :: TimeLocale
            , dfcDateFormat :: DateFormat
            }

I hope these changes make sense.

PR #459, opened 2016-08-03
optimise CSS compression

bergus:
I felt there's some long-hanging fruit :-)
In my project, it brought 10038 bytes down to 9752, which is not much, but the result did look much denser: only 284 whitespace characters left after compressing instead of 570.

Removing whitespaces next to `()!` characters might be the next step, but that probably doesn't make much difference. And for anything else, we'd have to leave the realm of simple text replacements.


jaspervdj:
There was a bug in CSS compression (#463) which was merged and conflicts with this.  I would be happy to merge this at a later point though.

bergus:
Shall I resolve the conflict via a merge, or should I do a rebase?

bergus:
Oh well, maybe I should've rebased it. Regardless, please check the new HEAD!

src/Hakyll/Web/CompressCss.hs
```diff
@@ -8,12 +8,14 @@ module Hakyll.Web.CompressCss
 
 
 --------------------------------------------------------------------------------
-import           Data.List               (isPrefixOf)
+import           Data.Char               (isSpace)
+import           Data.List               (dropWhileEnd, isPrefixOf)
```
bergus: I have no idea why but I'm getting a warning
```
The import of `isPrefixOf' from module `Data.List' is redundant
```

src/Hakyll/Web/CompressCss.hs
```diff
@@ -25,61 +27,51 @@ compressCssCompiler = fmap compressCss <$> getResourceString
...
-        | otherwise = eatComments $ drop 1 str'
+stripComments ""                       = ""
+stripComments ('/':'*':str)            = stripComments $ eatComment str
+stripComments (x:xs) | x `elem` "\"'"  = retainString x xs stripComments
+                     | otherwise       = x : stripComments xs
+
+eatComment :: String -> String
+eatComment "" = ""
+eatComment ('*':'/':str) = str
+eatComment (_:str) = eatComment str
```
bergus: Not sure whether
```hs
stripComments "" = ""
stripComments (str@x:xs)
    | x `elem` "\"'"        = x : retainString x xs stripComments
    | "/*" `isPrefixOf` str = stripComments $ eatComment $ drop 2 str
    | otherwise             = x : stripComments xs
  where
    eatComment "" = ""
    eatComment str' | "*/" `isPrefixOf str' = drop 2 str'
    eatComment (_:str') = eatComment str'
```
might be more idiomatic Haskell.

bergus:
OK, this could be merged now as I resolved the conflicts.
It's not much of a functional improvement after #545 any more, but I feel the code is still cleaner.

jaspervdj:
Thanks!

PR #462, opened 2016-08-04
Better error messages

### LINKED ISSUE #461: Better error messages
In #460 and in general before when doing complicated stuff in custom `fields`/`functionField`s I noticed that Hakyll always only reports the error message from `missingField`, and not the actual cause of any mistakes.
I even did resort to `return (error "…") :: Compiler String` at some points to get my error messages out to the user (me), which is a very ugly hack.

A quite useful improvement would be to change the `Monoid` instance of `Context`s to

``` haskell
    mappend (Context f) (Context g) = Context $ apply
      where
        apply k a i = f k a i `catchError` \errf ->
                        g k a i `catchError` \errg -> throwError (errf++errg)
```

so that they report all errors that happened when trying to evaluate a field.

I'm not sure though how many messages this might create if multiple fields are chained and whether that might become too many. If it is not too verbose, one might replace all the `empty` return values from contexts (currently it's equivalent to `throwError []`) with things like "_tried field xy_", "_could not generate route for yz_", "_failed to find field wx in metadata of uv_", "_field vw is false_" etc. The date field already does this when failing to parse. While the message would get possibly a bit long, it gives a good summary of what is available in the particular context.

Also it would be helpful if the `$if(…)$` evaluation could report errors. I understand this is a bit complicated as there is no distinction between "field threw error from computation" and "field does not exist". It would be awesome if "expected errors" (no metadata, wrong field name, false boolfields) could somehow be tagged and suppressed while the rest leads to a failing template compilation. This is a breaking change of course, and I'm not sure about its feasibility. Still, maybe you could output errors from if-conditions to `Logger.debug`.

Another great improvement for error messages would be if they reported the name of the template that failed to get applied. This can however not be done by simply adding some parameters to internal functions (like passing the `identifier` right through in `loadAndApplyTemplate`), but would likely require the addition of a `name` field to the `Template` constructor. Oh wait, `Template` implementation _is_ internal :-)

---

I'd like to hear your thoughts on this and can submit a pull request if you think it's worth it.
Having peeked around in the Hakyll source for some time now, it does seem rather easy; but from what I remember unhelpful error messages were nearly a showstopper when I began using Hakyll and writing fields basically had rendered templates as results…

### LINKED ISSUE #504: Template parser should fail or warn on unmatched `$`
A tricky-to-find bug was having an unterminated `$` in a template, which truncated the whole document after the first `$`. If we could build this into the parser or linter, it could make things easier for some people in the future.

### LINKED ISSUE #507: Invalid template control flow crashes silently
Stack Resolver lts-6.27
GHC 7.10.3
Hakyll 4.8.3.2
Mac OSX El Capitan 10.11.4

If you use the following in a template (terminating `for` with `endif`), the template renderer seems to just crash, rendering only the HTML up until the invalid syntax.

```html
$for(xs)$
  <p>foo</p>
$endif$
```

bergus:
Closes #461, closes #507, closes #504 (via 9ec43a6ed4)
- Better error messages for context fields
- Better error messages for template applications
- Better error messages for template parsing (there was a bug, `many` just dropped out in the middle of a template when it encountered invalid syntax)


src/Hakyll/Web/Template/Context.hs
```diff
@@ -157,7 +160,7 @@ snippetField :: Context String
 snippetField = functionField "snippet" f
   where
     f [contentsPath] _ = loadBody (fromFilePath contentsPath)
-    f _              i = error $
```
bergus: I suppose using `error` here was not on purpose?

src/Hakyll/Web/Template/Context.hs
```diff
@@ -198,15 +201,19 @@ bodyField key = field key $ return . itemBody
 -- | Map any field to its metadata value, if present
 metadataField :: Context a
 metadataField = Context $ \k _ i -> do
-    value <- getMetadataField (itemIdentifier i) k
```
bergus: I'm not sure whether to just use `getMetadataField'` directly

bergus:
Hm, with every Context field now creating a message when it does not apply that does create an awful lot of debugging messages from the Compiler Alternative. Is `<|>` typically used directly by Users? If no, it's only usage is in the Context monoid, and maybe we should drop the debug messages from there. Or at least filter out the ones that begin with "Tried field".

jaspervdj:
This is great work, but I'm not sure about the change to `Alternative`.

```
compilerCatch y $ compilerThrow . (es++)
```
If you know run in verbose mode (`-v`), this will print a quadratic amount of messages, and then the error at the end will repeat them all once more.  If we want to go this way, we probably want to remove the call to the logger there.

bergus:
Yes, that's why I initially thought to drop the `<|>` call from Context's `<>` and implement it directly. Do you know whether any Users are calling `<|>` directly from the applications?

What I also considered is to change the `CompilerError` constructor to `[(Verbosity, String)] -> CompilerResult a`. This could allow us to create results that take the second alternative path, but only contain debug messages not exceptions. Not sure how feasible that would be, but we could
* print only the actual error when there's a mistake that caused an exception
* print the debug messages when there is no real error, which would show all the attempted fields (ending with "missing")
* fail with an error when an `$if(…)$` field is "empty" because of an exception

The last one would be a breaking change, but possibly a good one; if not, we still could log these errors.
What do you think?

src/Hakyll/Web/Template/Internal.hs
```diff
@@ -157,11 +158,14 @@ applyTemplate' tes name context x = go tes `catchError` handler
...
+    applyElem (If e t mf) = do
+        c <- (applyExpr e >> return True) `compilerCatch` handler
+        if c
+            then go t
+            else maybe (return "") go mf
       where
-        handler _ = case mf of
-            Nothing -> return ""
-            Just f  -> go f
+        handler Error es = compilerThrow es
```
bergus: To match the old behaviour, this would need to be
```hs
        handler Error es = do 
            logger <- compilerLogger <$> compilerAsk
            forM_ es $ \e -> compilerUnsafeIO $ Logger.debug logger $
                "Hakyll.Web.Template.applyTemplate: Error in 'if': " ++ e
            return ""
```

src/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -164,25 +165,28 @@ instance MonadMetadata Compiler where
 --------------------------------------------------------------------------------
 instance MonadError [String] Compiler where
   throwError = compilerThrow
-  catchError = compilerCatch
+  catchError = (. matchErr) . compilerCatch
+    where
+      matchErr f Logger.Error es = f es
+      matchErr f _            _  = f []
```
bergus: I do wonder whether this is a good idea. If a compiler failed with some verbosity below `Error`, why give no information to the handler at all? It might be useful.
Still, the argument is that ```compile `catchError` throwError``` should behave more like identity, and not change the verbosity level.

bergus:
While working on this, I found another bug: When the body of an `$if(…)$` statement contained a field that threw an error, the `else` body ran. Ooops.
Now it correctly raised such errors.

bergus:
Please see the update. I've changed `CompilerError` to include a verbosity (not multiple messages of different verbosities, as I considered initially), and the `Alternative` does drop the less important messages if both compilers fail.
I have removed the logging in verbose mode completely, as fields now either fail with a loud error or not at all. Notice that the `-v` output even got shorter than before, as I removed the `missingField` from the `defaultContext` - the `applyTemplate'` does add a `missingField` to every context anyway - and previously that duplication would have triggered the debug log with the first message in the Alternative.

bergus:
Btw, when do you plan the next release? I'm close to shipping and need to decide whether to wait for an official version bump or just put my branches in as a beta.

src/Hakyll/Web/Template/Internal.hs
bergus: I've had a quick google search for `field` and `fail` usage scenarios, and found none that were offending. Or actually, [this one](//github.com/jaspervdj/hakyll/pull/348#issuecomment-126206137).

src/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -164,25 +165,28 @@ instance MonadMetadata Compiler where
...
-        logger <- compilerLogger <$> compilerAsk
-        forM_ es $ \e -> compilerUnsafeIO $ Logger.debug logger $
-            "Hakyll.Core.Compiler.Internal: Alternative failed: " ++ e
-        compilerCatch y $ compilerThrow . (es++)
+    empty   = compilerError Logger.Debug []
+    x <|> y = compilerCatch x $ \vx es -> compilerCatch y $ \vy ->
+        case vx `compare` vy of
+            LT -> compilerError vx . const es
+            EQ -> compilerError vx . (es++)
+            GT -> compilerError vy
```
bergus: I suppose that when dropping a list of messagse because there are more important ones one could log them to the debug screen. Not sure whether necessary.

bergus:
Another question I'd like to get some feedback on is the order/formatting of the error messages. I like to use things like
```hs
-- helper to amend error messages
withErrorContext :: MonadError [String] m => String -> m a -> m a
withErrorContext msg = flip catchError (throwError . (msg:))
```
which basically prepends a string to the list of error messages if an error happens. This does give us some kind of call stack.
When logging the list of messages, should it be `reverse`d so that the innermost error is at the top and the rest of lines gives the context? Or should the innermost error be printed as the last line?

jaspervdj:
Hey, this PR is really starting to look god, I appreciate all the effort!  Unfortunately I'm traveling to the states this weekend so I won't be able to look into it in detail until monday.  I hope to merge it it then.

bergus:
Thanks, I don't want to put any pressure on you :-)
I'm trying to use serverside builds now, and those don't work with a custom hakyll installation. Can you give me a rough estimate when we can expect hakyll-4.9 (or 5.0?) on hackage?
Otherwise I'll need to use [this approach](http://stackoverflow.com/questions/29153999/specify-a-github-url-for-a-dependency-in-a-cabal-file) to install hakyll from my branch, or just revert my files to be 4.8.3 compatible.

jaspervdj:
I would prefer to keep the `$if(..)$` behavior backwards-compatible.  It is fairly common in Haskell libraries that `empty` is equivalent to `fail msg` apart from the produced message.  Other than that, everything looks good to merge.

We can probably do a release later this week.  If you're having trouble using custom versions of libraries, you can always consider using [stack](https://docs.haskellstack.org/en/stable/README/).  I need to use patched libraries all the time at work and I've never had any trouble with it.

bergus:
Yeah, I'm right in the process of switching from a plain `ghc` invocation to `stack` :-)

I'm going to change the `$if$` behaviour back and instead log errors to the debug view.
But regarding b/c, notice that 8ac7949eaa90 also breaks the `Template` constructor (which is no longer exported from `Template.Internal`) and replaces it by the `template` "smart constructor".

bergus:
Hi, have you had a chance to look at this?

jaspervdj:
Hey, I just had a look.  I think the code is kind of confusing because the `Verbosity` has a bunch of different effects.  I think it would be cleaner if I would refactor it all a bit to have some sort of `Writer`-like monoid in `Compiler` for logging?  That way, we can just accumulate all the messages cleanly, and then at the end decide locally which ones to print etc.  What do you think?

bergus:
You mean running the `Writer` to get its results at places where you actually want to log something?
I had thought of doing `CompilerError :: [(Verbosity, String)] -> CompilerResult a`, but then I did like the current solution better because in the end we really only want to get the errors of most prominence anyway.
But sure, feel free to clone my branch and refactor it to show me what you mean.

bergus:
This is still open. If you want me to solve the conflicts so that you can merge, please comment.

bgamari:
I, for one, would appreciate better error messages. I've fought some bitter battles with Hakyll due to its at times unhelpful error messages.

jaspervdj:
Sorry for the delayed response on my part.

I'm still very interested in getting a form of this merged in, but I'm not sure when I'll have time to look at it. :-(

It currently seems that we could solve this by putting something like `[(Verbosity, String)]` in `CompilerWrite`, and modifying the `CompilerError` constructor to `:: CompilerWrite -> CompilerResult a`.

bergus:
OK, I'll try to get this working again after resolving the conflicts, and then see what I can do.
It's been a long time, which is a good way to evaluate whether I wrote maintainable code that I can still understand :-)

bergus:
OK, I've reviewed this `Verbosity` thing. I guess using `Logger.Verbosity` is confusing indeed, as it doesn't directly have to do with logging.

All this hassle arises from the design of `Template.Context` which uses the `Alternative` instance of `Compiler`s to "try" one field after the other and deciding `$if(…)$`s.
We would actually like to distinguish exceptions (IO errors, syntax errors, type errors, Haskell exceptions) from fields that could not be applied. However, currently both use `CompilerError` for signalling and collecting messages.
I can see two proper solutions for this:
* Introducing a new `CompilerNoResult :: [String] -> CompilerResult a` constructor which can be treated separately from errors in templates. It would get appropriate handling in the `Alternative Compiler` instance and might even be provided as the `empty :: Compiler`.
  Problem: No backcompat with code that uses `fail` in custom `Context`s.
* Introducing a new `MissingField :: ContextField` constructor that can be returned in contexts to signify that nothing was found. Or similarly, changing the return type of the `Context` functions to `Compiler (Maybe ContextField)`. Or rather, using `Either [String]` instead of `Maybe`, or respectively giving `MissingField` a `[String]` argument, so that we can collect helpful messages in there. (Directly logging them in the successfull compiler is not appropriate).
  Problem: No backcompat with code that uses `fail` or `empty :: Compiler` in custom `Context`s.

If you want to throw the backcompat concerns over board and release it as Hakyll 5.0, I'm all in :-)

My current approach resembles the first idea. Instead of using a separate constructor, I added a level to the `CompilerError` - either `Message` or `Error` (`Debug` isn't really used). The `Alternative` instance is a bit weird, but it basically says that when both attempts fail an error will take precedence over a normal message. If both failures are of the same level, their messages are concatenated.

For backcompat, even when the first alternative fails with an exception, the second one is still tried. In a proper solution, an exception would not be ignored - at least for the purposes of templates, not sure whether an error-suppressing alternative is useful elsewhere or whether `catchError` is enough for that.

> some sort of `Writer`-like monoid in `Compiler` for logging? That way, we can just accumulate all the messages cleanly, and then at the end decide locally which ones to print etc. What do you think?

"at the end" sounds like "globally" to me? And I'm not sure whether this will work, filtering for level/verbosity in the end - we would want to locally decide which messages to keep. The template evaluation forms a tree, and if we just accumulate we would need to somehow represent that structure in the monoid?
Or maybe I just don't really understand what one can do with `Writer`, I have never used it before.

> It currently seems that we could solve this by putting something like `[(Verbosity, String)]` in `CompilerWrite`, and modifying the `CompilerError` constructor to `:: CompilerWrite -> CompilerResult a`.

Not sure whether I grasp the scope of that change. Why use `CompilerWrite` here? What would the list of messages do in a `CompilerDone`? What would the dependencies and cache hits do in a `CompilerError`? I don't feel comfortable doing such an edit.

I will try to remove the `Verbosity`. Either by implementing `data Reason = Error [String] | Message [String]` and having `CompilerError :: Reason -> CompilerResult a`, or with that `CompilerNoResult` - which I quite like actually, even despite the backcompat `fail` semantics.

bergus:
OK I tried
```hs
data CompilerResult a
    = CompilerDone a CompilerWrite
    | CompilerSnapshot Snapshot (Compiler a)
    | CompilerRequire (Identifier, Snapshot) (Compiler a)
    | CompilerNothing [String]
    | CompilerError [String]
```
but it's still really ugly:
```hs
compilerResult x = Compiler $ \_ -> return x
compilerThrow = compilerResult . CompilerError
compilerNothing = compilerResult . CompilerNothing

compilerHandle :: ([String] -> Compiler a) -> ([String] -> Compiler a) -> Compiler a -> Compiler a
compilerHandle f g (Compiler x) = Compiler $ \r -> do
    res <- x r
    case res of
        CompilerDone res' w  -> return (CompilerDone res' w)
        CompilerSnapshot s c -> return (CompilerSnapshot s (compilerHandle f g c))
        CompilerRequire i c  -> return (CompilerRequire i (compilerHandle f g c))
        CompilerNothing e    -> unCompiler (g e) r
        CompilerError e      -> unCompiler (f e) r

instance Alternative Compiler where
    empty   = compilerNothing []
    x <|> y = compilerHandle (\rxs ->
          compilerHandle (\rys ->
            compilerNothing $ rxs ++ rys
          ) (\eys ->
            log rxs >> compilerThrow eys
          ) y
        ) (\exs ->
          compilerHandle (\rys ->
            log rys >> compilerThrow exs
          ) (\eys ->
            compilerNothing $ exs ++ eys
          ) y
        ) x
      where
        log = compilerDebugLog . map
            ("Hakyll.Core.Compiler.Internal: Alternative fail suppressed: " ++)
```

lib/Hakyll/Web/Template.hs
```diff
@@ -140,14 +140,13 @@
 --
 module Hakyll.Web.Template
     ( Template
-    , template
-    , readTemplateElems
```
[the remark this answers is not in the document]
jaspervdj: Yeah, we should be good to rip out `unsafeReadTemplateFile`

jaspervdj:
I would also favor the first approach you presented, and the code doesn't look too bad IMO.

But I'm a bit unsure about the `Alternative` instance -- it looks like `x <|> empty` might change `CompilerError` to `CompilerNothing` or the other way around?  Wouldn't that violate the laws?

I also think `CompilerEmpty` is a better name than `CompilerNothing`.

As far as backwards compatibility goes, I care most about keeping everything working.  If users get a slightly less good error message because they are using e.g. `fail` rather than `empty` (or something else entirely), that's fine.

bergus:
> code doesn't look too bad IMO.

I did like the `Alternative` implementation with the `Verbosity` much better than this `compilerHandle` thingy with two callbacks. Also `CompilerNothing` needed to be added almost everywhere as a repetition of the `CompilerError` case. Admittedly `Verbosity` was a horrible choice, I went with a new data type designed specifically for this now. See commit below.

> But I'm a bit unsure about the Alternative instance -- it looks like `x <|> empty` might change CompilerError to CompilerNothing or the other way around? Wouldn't that violate the laws?

The idea is that `CompilerError` always takes precedence over `CompilerNothing`. `x <|> empty = x` should hold - the `Nothing []` gets dropped, and there shouldn't even be a debug log entry from no messages.

> I also think CompilerEmpty is a better name than CompilerNothing.

That name I wasn't happy about either :-) Good suggestion - I had the same idea for renaming the `NoField`.

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -90,11 +94,25 @@ instance Monoid CompilerWrite where
...
-    CompilerError    :: [String] -> CompilerResult a
-    CompilerRequire  :: (Identifier, Snapshot) -> Compiler a -> CompilerResult a
+data Reason a
+    -- | An exception occured during compilation
+    = CompilationFailure a
+    -- | Absence of any result, most notably in template contexts
+    | NoCompilationResult a
+
+
+-- | Unwrap a `Reason`
```
bergus: I guess `unReason` might have followed the naming conventions, but it seemed kinda unreasonable :-D

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -166,30 +182,29 @@ instance MonadMetadata Compiler where
 --------------------------------------------------------------------------------
 instance MonadError [String] Compiler where
   throwError = compilerThrow
-  catchError = (. matchErr) . compilerCatch
-    where
-      matchErr f Logger.Error es = f es
-      matchErr f _            _  = f []
```
bergus: I don't remember why I did pass no messages to the the handler here - to make it distinguishable?
Now `catchError` will also catch `NoCompilationResult`s, not just `CompilationFailure`s, which seems reasonable - they are errors, at least outside of template applications.
However, ``c `catchError` throwError`` will change the error reason type, which I am not so happy about. Maybe we should export a helper function that allows altering (adding to) the messages without changing the type. [`Bifunctor`](https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Bifunctor.html) comes to my mind, but our messages are string lists not arbitrary types.

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -166,30 +182,29 @@ instance MonadMetadata Compiler where
...
-            LT -> log eys >> compilerError vx exs
-            EQ ->            compilerError vx (exs ++ eys)
-            GT -> log exs >> compilerError vy eys
+    empty   = compilerMissing []
+    x <|> y = x `compilerCatch` (\rx -> y `compilerCatch` (\ry ->
+        case (rx, ry) of
+          (CompilationFailure xs,  CompilationFailure ys)  -> compilerThrow $ xs ++ ys
+          (CompilationFailure xs,  NoCompilationResult ys) -> debug ys >> compilerThrow xs
+          (NoCompilationResult xs, CompilationFailure ys)  -> debug xs >> compilerThrow ys
+          (NoCompilationResult xs, NoCompilationResult ys) -> compilerMissing $ xs ++ ys
```
bergus: I could have made `Reason` a `Semigroup` or `Monoid` instance for this to make the code even cleaner, but that doesn't allow to put the discarded messages in the `debug`.

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -18,7 +18,7 @@ module Hakyll.Core.Compiler.Internal
     , compilerTell
     , compilerAsk
     , compilerThrow
-    , compilerFailMessage
+    , compilerNoResult
```
bergus: I guess this one should also be re-exported from `Core.Compiler`, as it's useful for people writing custom `Context`s with error messages.

lib/Hakyll/Core/Compiler/Internal.hs
jaspervdj: :grin:

jaspervdj:
The separate datatype makes the code much more obvious, nice job!

I think `compilerNoResult` needs a Haddock to indicate why it's there.  Should we also do `fail = compilerNoResult`?  I'm not fully convinced about the name yet.  Maybe something like `compilerFailBranch`?

Other than that, I think all the code looks good now! :-)

bergus:
> Should we also do `fail = compilerNoResult`?

No, `fail` should stay reserved for unrecoverable errors.

> I'm not fully convinced about the name yet. Maybe something like `compilerFailBranch`?

Me neither. `compilerFailBranch` sounds good! Haddock documentation coming.

jaspervdj:
This was merged in #722

PR #463, opened 2016-08-14
Bugfix: CompressCss modified CSS string constants which it should not do

NicoleRauch:
CompressCss applied compressions (e.g. removal of whitespace) to CSS string constants. But those constants should retain their contents in the compressed CSS.

I've added some tests to demonstrate the behaviour, and I fixed the code. My new code is still rather verbose, which I think can be improved. I'm open to suggestions!

### DIFF
--- a/hakyll.cabal
+++ b/hakyll.cabal
@@ -223,6 +223,7 @@ Test-suite hakyll-tests
     Hakyll.Core.Store.Tests
     Hakyll.Core.UnixFilter.Tests
     Hakyll.Core.Util.String.Tests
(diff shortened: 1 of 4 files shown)


NicoleRauch:
Hi Jasper,

did you have the opportunity to look at this pull request? Would you like me to improve something?

I'm quite keen on being able to use the standard release of hakyll again for my site (which this bug is blocking).

Cheers,
Nicole

jaspervdj:
Thanks, I'm sorry for the long time it took to review this!

PR #464, opened 2016-08-17
Fix feed generator when item contains CDATA

yogsototh:
Fix a bug when renderAtom and renderRss generate broken XML when the item contains CDATA blocks.


jaspervdj:
Thanks!

PR #469, opened 2016-08-31
prettyPrintParseException was introduced in yaml-0.8.11

juhp:
hakyll does not build with earlier versions of yaml that didn't provide `prettyPrintParseException`.


juhp:
Or is there a simple cpp workaround for this say?

jaspervdj:
I think it's fair to require the version bump.  Thanks for the PR!

PR #470, opened 2016-08-31
Remove https://dikmax.name/ from examples.

dikmax:
I don't use Hakyll anymore.

### DIFF
--- a/web/examples.markdown
+++ b/web/examples.markdown
@@ -66,8 +66,6 @@ directly with the default Hakyll site.
   [literate source](http://scr.stunts.hu/hakyll.html)
 - <http://www.eanalytica.com/>,
   [literate source](http://www.eanalytica.com/site/)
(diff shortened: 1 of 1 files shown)


jaspervdj:
Thanks for updating!  Let me know if there are any particular things you didn't like about Hakyll.

dikmax:
Hakyll is perfectly fine, and I liked it much.

I just rewrote everything with Shake to have more control and speed.
As a tradeoff, I had to write more code. 

And with Shake is easier to get rid of tons of bash files I wrote for additional processing.

listx:
So is it accurate to say that you wrote your own static site generator with
Shake?

dikmax:
@listx I think yes. If you interested in implementation, you can check source at [dikmax/dikmax.name-ng](https://github.com/dikmax/dikmax.name-ng).

PR #472, opened 2016-09-06
Update git fetch and fix rsync for stack

jakzale:
There are two changes:

1. Fixed `git fetch` to use `--all` flag
2. Added `.stack-work` to directories ignored by rsync, otherwise `.stack-work` gets deleted by rsync (Ouch!).  I also refactored the rsync command to be more readable.


jaspervdj:
Thanks!  In fact, this tutorial should be rewritten to make use of the recent github pages option to pull a static site from the `/docs` directory.  That makes things a lot easier.  I need to see if I can do that at some point.

PR #482, opened 2016-10-31
Fix integer fields in YAML metadata

nickie:
If a numeric field contains an integer number, it should be shown
as an integer number.  Without this patch, the field:

    answer = 42

would be shown as "42.0".  Before the introduction of YAML in
metadata, this was treated as a string and therefore shown as
just "42".  This patch reinstates the older (and IMHO correct)
behavior.


jaspervdj:
Thanks!  Patch looks good to me, but could you add a comment in the source code about why this conversion is a good thing (basically the description in your PR)?  That way I won't wonder what is happening there later on.

nickie:
Done.  I also added a test for the intended behavior.  You may want to squash-merge these three commits.

jaspervdj:
Thanks!

PR #484, opened 2016-11-13
Added alexanderlobov.ru to examples

alexanderlobov:
Hi, Jasper!
I have added alexanderlobov.ru an an example of site using Hakyll.


jaspervdj:
Thanks!

PR #485, opened 2016-11-18
Enable bypassing of argument parsing

### LINKED ISSUE #400: Bypassing argument parsing
As Hakyll.Commands and Hakyll.Core.Runtime are not exported packages, there doesn't seem to be a way to write a main function that bypasses Hakyll's command line parsing (say, to provide a different command line interface.)

My use case is the integration of Hakyll's site building process in an existing make-based environment in a separate build tree. I would like not to have to modify the site.hs source to modify source and destination directories in-place, but instead provide these as command-line parameters.

An alternative to exporting these packages would be to provide functions in Hakyll.Main that split the responsibility of parsing the command-line arguments from invoking the Hakyll commands.

(edit) Rather belatedly it occurs to me one can use withArgs to work-around this issue, though it does lack a certain elegance.

aesadde:
Closes #400.

Two functions added:
  * 'hakyllWithArgs' exported which expects the command line arguments
    to be given by the user.
  * 'hakyllWithExitCodeAndArgs' used internally selects the relevant
     arguments and invokes the commands.

### DIFF
--- a/src/Hakyll/Main.hs
+++ b/src/Hakyll/Main.hs
@@ -1,10 +1,11 @@
 --------------------------------------------------------------------------------
 -- | Module providing the main hakyll function and command-line argument parsing
 {-# LANGUAGE CPP                #-}
-{-# LANGUAGE DeriveDataTypeable #-}
+
 module Hakyll.Main
     ( hakyll
     , hakyllWith
+    , hakyllWithArgs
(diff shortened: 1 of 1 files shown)


jaspervdj:
Thanks for the contribution, and sorry for the latency!

PR #501, opened 2017-01-04
Fail check on non success status code

clample:
Currently, external links resulting in a 404 error or other non success status codes will still pass the check.

With these changes, the user is now alerted about non success status codes from external links.

### DIFF
--- a/src/Hakyll/Check.hs
+++ b/src/Hakyll/Check.hs
@@ -242,9 +242,14 @@ checkExternalUrl url = do
             case (cast e :: Maybe SomeAsyncException) of
                 Just ae -> throw ae
                 _ -> faulty url (Just $ showException e)
-        Right _ -> ok url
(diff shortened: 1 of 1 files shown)


clample:
I'm not sure if an error makes sense for all non 2** response codes. It seems like we might get non 2** response codes for pages that are still accessible from the browser. For example, `https://medium.freecodecamp.com/yes-im-56-and-learning-to-code-f33abea6fd4c#.zhqs1isvb` gave a 409 error when running the check tool.

On the other hand, it might be better to over-report issues when looking for broken links than to let possible issues go unnoticed.

jaspervdj:
This looks good but I think that `3xx` status codes should also be valid, external URLs should be free to be moved around.

LaurentRDC:
Done in 8170110c53601e5135c2988f65e12e92d6bdfb97

PR #506, opened 2017-01-05
Make url check concurrent

### LINKED ISSUE #479: Link checker: make requests in parallel
Given the amount of latency involved in sending a request, it would be nice if the checker performed its requests in parallel.

clample:
Resolves #479

### DIFF
--- a/src/Hakyll/Check.hs
+++ b/src/Hakyll/Check.hs
@@ -8,22 +8,21 @@ module Hakyll.Check
 
 
 --------------------------------------------------------------------------------
-import           Control.Monad                (forM_)
-import           Control.Monad.Reader         (ask)
-import           Control.Monad.RWS            (RWST, runRWST)
+import           Control.Monad                (forM_, foldM)
+import           Control.Monad.Reader         (ask, ReaderT, runReaderT)
 import           Control.Monad.Trans          (liftIO)
 import           Control.Monad.Trans.Resource (runResourceT)
-import           Control.Monad.Writer         (tell)
 import           Data.ByteString.Char8        (unpack)
 import           Data.List                    (isPrefixOf)
-import           Data.Set                     (Set)
-import qualified Data.Set                     as S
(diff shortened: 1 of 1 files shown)


clample:
I built the example site and added 50 external links from hacker news: [Sample Markdown](https://gist.github.com/clample/00283ed45ed4ac3579e36b481491b6fc)

I ran the link checker against the sample site to see if these changes actually improved performance. I ran the link check 10 times with the concurrent link check and 10 times with master. The average time improved from 27.2 seconds to 20.5 seconds.

Actual results:

With concurrency:

13.882
13.807
16.456
13.639
14.385
15.356
20.795
28.657
40.384
27.241

average: 20.4602

Master (without concurrency):

17.040
30.740
29.492
32.440
27.697
25.157
26.285
30.854
23.844
28.244

average: 27.1793

clample:
I might also have time to add tests (we'll see)

src/Hakyll/Check.hs
```diff
@@ -191,58 +233,47 @@ checkInternalUrl base url = case url' of
...
+                _ -> faulty url (Just $ showException e)
+        Right _ -> ok url
+    where
+        -- Convert exception to a concise form
+        showException e = case cast e of
+            Just (Http.HttpExceptionRequest _ e') -> show e'
+            _ -> head $ words $ show e
 #else
-checkExternalUrl _ = return ()
+checkExternalUrl url = skip url Nothing
```
clample: We do `skip` to make sure we eventually write to the MVar. Otherwise I believe that there's an error

jaspervdj:
Looks good to me 👍 (although I think this could be simplified a bit by just using an `IORef` instead of an `MVar`).

Can I merge this in or do you want to add additional work?

clample:
Cool! (I didn't know what an `IORef` is but I'll check it out).

You're welcome to merge this

PR #512, opened 2017-01-17
Added information about the separator functionality in $for$.

beerendlauwers:
This information can only be found in the Haddocks documentation, but it's something you use often, so I've added it here.


jaspervdj:
Thanks!

PR #520, opened 2017-02-10
Preserve file metadata in copyFileCompiler

frederik-h:
I think copyFileCompiler should preserve the metadata of files, in particular the mtime.
This avoids unnecessary file transfers when deploying, e.g., with rsync.

(Note: copyFileWithMetadata has been introduced to directory rather recently, in version 1.2.6.0 from April 2016)

### DIFF
--- a/src/Hakyll/Core/File.hs
+++ b/src/Hakyll/Core/File.hs
@@ -14,7 +14,7 @@ module Hakyll.Core.File
 --------------------------------------------------------------------------------
 import           Data.Binary                   (Binary (..))
 import           Data.Typeable                 (Typeable)
-#if !MIN_VERSION_directory(1,2,6)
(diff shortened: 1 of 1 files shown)


jaspervdj:
`hakyll.cabal` specifies that we are compatible with `directory 1.0`.  Would it be possible to wrap the new definition in an `#ifdef MIN_VERSION_directory` preprocessor statement, and use the old `copyFile` if the directory version is older than 1.2.6?

frederik-h:
I added the #if 
(but messed up the commit history badly ...)

jaspervdj:
Don't worry, I will squash-merge it.  I think you want to delete the `!` though, it looks like the branches of the if should be the other way around.

frederik-h:
Fixed (and tested)

jaspervdj:
Thanks!

PR #530, opened 2017-03-16
Remove `argumatronic.org` from examples.

misandrist:
This woman is a well-known supporter of neo-Nazi ideologies, and has
expressed quite a willingness to shoot anyone who disagrees with
her. I don't know that she's a great example of someone a project
should be associated with.

For example, this thread where she and a friend support buying ammo
for any FPer because, well, obviously they believe all FPers are Nazis
who might get punched or something:

ammo-for-fp

argumatronic-006

And of course she proudly contributed to giving yet another platform
to White Supremacy. And, as we've seen from the uptick in
violence (more like a groundswell) by White Supremacists even before
the U.S. election, I think that there's really no question that paying
for someone with that ideology to speak is the same kind of enablement
that makes the truly violent people feel much more comfortable in
exercising that violence, which she and hers already have expressed a
willingness to do in order to defend their right to support violence.

### DIFF
--- a/web/examples.markdown
+++ b/web/examples.markdown
@@ -168,8 +168,6 @@ directly with the default Hakyll site.
   [source](https://github.com/beerendlauwers/HaskAnything)
 - <https://www.brandonstil.es/>,
   [source](https://github.com/stilesb/brandonstil.es)
-- <http://argumatronic.com/>,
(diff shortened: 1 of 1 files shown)


jaspervdj:
I know this woman and she is not "a well-known supporter of neo-Nazi ideologies".

PR #534, opened 2017-04-11
Add tutorials

alx741:
So I just wrote this two tutorials:
- Port an existing Blogger blog to Hakyll
- Using Cassius template with Hakyll


jaspervdj:
Thanks!

PR #536, opened 2017-04-17
Add Semigroup instance for Context

sergv:


jaspervdj:
@sergv This looks good to me but it's a bit weird that we're checking the version of `directory` rather than `base`.  Is there any reason for this?

sergv:
@jaspervdj Definitely, I should have specified `base` instead of `directory`.

PR #541, opened 2017-06-09
Tighten the dependency on pandoc-citeproc.

23Skidoo:
0.10.5 added a new type `CLabel` to `Text.CSL.Reference`, which broke the orphan instance in `Web.Pandoc.Binary`.

Note also that according to PVP a client that defines an orphan instance **must** depend on the *minor* version of the package defining the type.


23Skidoo:
Would be also nice if you could update the bounds on Hackage.

jaspervdj:
Thanks!

Issue #140, opened 2013-05-03
Implement live-reload for preview mode


meiersi:
A recent trend for web development is to use live-reload to get rid of the need to refresh the browser by yourself after changing the source files. The implementations work by injecting a piece of JavaScript into the page, which then connects to your preview server and forces a reload if told so. One implementation can be found here: https://github.com/gruntjs/grunt-contrib-livereload.

It should be rather simple to adapt https://github.com/gruntjs/grunt-contrib-livereload to be used in the preview mode of Hakyll.

benkolera:
For anyone that comes along here later. The easiest way that I found was just to reuse the existing nodejs library. It's easy if you already have npm. I wrote about it here: http://benkolera.com/posts/2015-09-14-hakyll_livereload.html

Given that most web dev people tend to have nodejs around for building web stuff and the live reload server is super simple and works, I don't see an intense need to build something into site watch that implemented the LiveReload protocol or anything like that.

Could be documented on the hakyll wiki, though. I'm happy to submit a PR with my docs if that seems like it'd be useful to someone else! :smile:

jaspervdj:
That's a nice and simple solution @benkolera! Feel free to send a PR which adds a tutorial to the Hakyll website.

Wizek:
Nice idea about LiveRealod! Webarchive link for the 404-ing page: 

https://web.archive.org/web/20160414044810/http://benkolera.com/posts/2015-09-14-hakyll_livereload.html

haoyun:
Recently I found live-reload is really useful when building a website, and realized that Hakyll watchserver does not have this great feature. I tried to search the internet and found out this old issue. Besides the above mentioned ideas or solutions, I also found 
* https://www.blaenkdenum.com/posts/live-editing-with-hakyll/
* https://github.com/iand675/live-reload
I started to try to implement it following the ideas above, but finally I found a much simpler solution:
* http://livejs.com/

Just include a JS then problem is (kind of) solved.
Hope this will help other people.

manveru:
Little script for people still looking for a one-liner to recompile the site and clean/watch. This works well combined with the livejs script. The watching is done using http://entrproject.org but you could probably also use https://github.com/watchexec/watchexec

```bash
echo site.hs | entr -n -s 'ghc --make site.hs && ( echo site | entr -n -r -s "./site clean && ./site watch" )'
```

PR #542, opened 2017-06-09
Expose settings for the preview server

league:
This enables customization of the settings used by the preview server,
for example to implement URL rewriting.

My particular use case is to mimic the `tryFiles` setting in nginx to
add an extension when looking for a matching file. Then the URL can be
`/projects/web` and it will match a file called `projects/web.html` --
a nice clean URL that doesn't require lots of subdirectories and every
file being called `index.html`.

(This change doesn't implement that feature directly as part of
Hakyll, it just exposes the settings to enable building features like
it in your `site.hs`.)

### DIFF
--- a/src/Hakyll/Commands.hs
+++ b/src/Hakyll/Commands.hs
@@ -120,8 +120,8 @@ rebuild conf logger rules =
 server :: Configuration -> Logger -> String -> Int -> IO ()
 #ifdef PREVIEW_SERVER
 server conf logger host port = do
-    let destination = destinationDirectory conf
-    staticServer logger destination host port
+    let settings = previewSettings conf $ destinationDirectory conf
+    staticServer logger settings host port
 #else
 server _ _ _ _ = previewServerDisabled
 #endif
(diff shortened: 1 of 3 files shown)


puffnfresh (approved):
This functionality would be useful to me for the same reason.

The change is very straightforward. I'll fix the conflicts and run this patch locally but I would love to see it merged.

jaspervdj (approved):

puffnfresh:
Conflict fixed here: https://github.com/puffnfresh/hakyll/commit/492effda9d5fade5960d8f1aa4f7ab0676400f8d

jaspervdj:
Merged manually in a7e7e52302fd38130ac5ceb677d81bff82af45d6

jtojnar:
This is great. It also [allows transforming the preview server response](https://gitlab.com/tojnar.cz/hakyll-contrib-tojnar/-/merge_requests/1), which can make absolute URLs of the site (e.g. in RSS feeds) work during testing. Plus it is useful for injecting JavaScript files for development like [live reloading](https://github.com/jaspervdj/hakyll/issues/140#issuecomment-382581799).

Issue #540, opened 2017-06-06
Doesn't build with pandoc-citeproc 0.10.5.1


felixonmars:
```
[ 5 of 52] Compiling Hakyll.Web.Pandoc.Binary ( src/Hakyll/Web/Pandoc/Binary.hs, dist/build/Hakyll/Web/Pandoc/Binary.o )

src/Hakyll/Web/Pandoc/Binary.hs:31:10: error:
    • No instance for (Binary REF.CLabel)
        arising from a use of ‘binary-0.8.3.0:Data.Binary.Class.$dmput’
    • In the expression: binary-0.8.3.0:Data.Binary.Class.$dmput
      In an equation for ‘put’:
          put = binary-0.8.3.0:Data.Binary.Class.$dmput
      In the instance declaration for ‘Binary REF.Reference’

src/Hakyll/Web/Pandoc/Binary.hs:31:10: error:
    • No instance for (Binary REF.CLabel)
        arising from a use of ‘binary-0.8.3.0:Data.Binary.Class.$dmget’
    • In the expression: binary-0.8.3.0:Data.Binary.Class.$dmget
      In an equation for ‘get’:
          get = binary-0.8.3.0:Data.Binary.Class.$dmget
      In the instance declaration for ‘Binary REF.Reference’
```

felixonmars:
Fix already landed. Thanks.

PR #543, opened 2017-06-09
fix for #540

nek0:
Just my little proposition to solve the errors arising whilst compilation.


jaspervdj:
Sorry, didn't notice the PR and fixed it already.  Thanks for the submission though.  :-)

PR #545, opened 2017-06-10
Some small CSS compression improvements

NicoleRauch:
I had these sitting on my computer for ages and almost forgot about them, so here they finally are.


jaspervdj:
Thanks!

PR #546, opened 2017-06-11
Restructure .cabal to avoid redundant compilation

league:
I noticed that when I make a change to one of the files in `src/`, it
gets compiled 2-3 times: once for the library, again for the
executable, and finally for the test suite.

There is a way to avoid this duplication: more cleanly segregate the
sources for each component: `src/` *only* for the library,
`tests/` *only* for the test suite, and `Init.hs` (moved out of
`src/`) for the executable.

The slight downside is that some of the 'hidden' modules in `src/`
have to be added to the `exposed-modules` of the library. (There was
already precedent for having some `.Internal` modules exposed --
naming them such might be deterrent enough to avoid external packages
from relying on them.)

Also, AFAICT there's no need for `other-modules` in a test-suite or
executable component.

You can decide if you agree with this trade-off, feel free to close without merging if you don't. Thanks!


jaspervdj:
Thanks!  I refactored it a bit more.

PR #553, opened 2017-07-09
Expose hakyllWithExitCodeAndArgs, Options, and Command.

### LINKED ISSUE #552: Options is not exported, making hakyllWithArgs useless
I'd like to pass some configuration into my Hakyll program, which makes the standard argument parsing unsuitable.  `hakyllWithArgs` looks like what I need, except the `Options` and `Command` types aren't exported, so there's no way to actually use it!

To be totally clear, here's what I envision my eventual code to be:

```haskell
main :: IO ()
main = do
  (myConfig, hakyllOptions) <- myArgumentParser =<< getArgs
  hakyllWithArgs defaultConfiguration hakyllOptions $ do
    ...
    -- something with myConfig
    ...
```

Also, while looking at `Main.hs`, I notice `hakyllWithExitCodeAndArgs` isn't exported either.  Maybe it should be?

barrucadu:
Fixes #552

### DIFF
--- a/lib/Hakyll/Main.hs
+++ b/lib/Hakyll/Main.hs
@@ -97,19 +97,31 @@ invokeCommands args conf check logger rules =
 
 --------------------------------------------------------------------------------
 
+-- | The parsed command-line options.
 data Options = Options {verbosity :: Bool, optCommand :: Command}
     deriving (Show)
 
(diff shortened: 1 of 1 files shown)


jaspervdj:
Thanks!

PR #557, opened 2017-07-21
First pass at Pandoc 2.0

rlpowell:
Based on whatever happened to be in Pandoc master branch as of ~15 July
2017.  I did this for myself so I figured I might as well share.

Note that I consider myself a mediocre Haskeller at best, so please
review this carefully.  It does appear to *work*, though, based on my
test suite for my blog.

### DIFF
--- a/lib/Hakyll/Web/Pandoc.hs
+++ b/lib/Hakyll/Web/Pandoc.hs
@@ -25,6 +25,8 @@ module Hakyll.Web.Pandoc
 import qualified Data.Set                   as S
 import           Text.Pandoc
 import           Text.Pandoc.Error          (PandocError (..))
+import           Text.Pandoc.Highlighting   (pygments)
+import qualified Data.Text                  as T
 
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lib/Hakyll/Web/Pandoc.hs
```diff
@@ -84,7 +86,10 @@ writePandoc = writePandocWith defaultHakyllWriterOptions
 writePandocWith :: WriterOptions  -- ^ Writer options for pandoc
                 -> Item Pandoc    -- ^ Document to write
                 -> Item String    -- ^ Resulting HTML
-writePandocWith wopt = fmap $ writeHtmlString wopt
+writePandocWith wopt (Item itemi doc) =
+    case runPure $ writeHtml5String wopt doc of
+        Left (PandocSomeError err)  -> Item itemi $ "Hakyll.Web.Pandoc.writePandocWith: unknown error: " ++ err
```
Balletie: Maybe use `fail $ "Hakyll.Web.Pandoc.writePandocWith: unknown error: " ++ err`?

vaibhavsagar: I tried this, but `Item` is not a `Monad` and doesn't have an implementation of `fail`.

PR #561, opened 2017-08-16
Add spyhop.cafe to example sites

thunky-monk:
Hey, thanks for maintaining Hakyll :)

I'd like to add this site to the list of examples. There's nothing notable about this site except that it doesn't use Hakyll templating. Instead it uses [Lucid](https://hackage.haskell.org/package/lucid) to define the HTML and passes along the metadata in a record type.

Cheers,
Aaron


jaspervdj:
Nice, thanks!

PR #562, opened 2017-08-17
Allow time-1.8, process-1.6 and pandoc-citeproc-0.11

felixonmars:
Builds fine and works here.


lfairy:
Arch Linux [does the same thing](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/haskell-hakyll&id=1112622ac74962e542690b9428c1c482f409ba7b), so that's another data point in favor of this change 👍

PR #565, opened 2017-09-02
Add boolFieldM which allows using the Compiler monad

### LINKED ISSUE #564: boolField with (Item a -> Compiler Bool)
The current `boolField` only accepts a function of type `Item a -> Bool`. In some cases it would be useful to have a version of `boolField` that accepts a function fo type `Item a -> Compiler a`.

Kritzefitz:
This commits adds the helper boolFieldM. In contrast to boolField it
allows using the Compiler monad in the passed function.
boolField is also redefined in terms of boolFieldM.
Closes #564

### DIFF
--- a/lib/Hakyll/Web/Template/Context.hs
+++ b/lib/Hakyll/Web/Template/Context.hs
@@ -6,6 +6,7 @@ module Hakyll.Web.Template.Context
     , Context (..)
     , field
     , boolField
+    , boolFieldM
     , constField
     , listField
     , listFieldWith
@@ -104,9 +105,17 @@ boolField
     :: String
     -> (Item a -> Bool)
     -> Context a
-boolField name f = field name (\i -> if f i
-    then pure (error $ unwords ["no string value for bool field:",name])
-    else empty)
+boolField name f = boolFieldM name $ pure . f
+
+--------------------------------------------------------------------------------
+-- | Like 'boolField' but allows the use of the 'Compiler' monad.
+boolFieldM :: String -> (Item a -> Compiler Bool) -> Context a
+boolFieldM name f = field name $ \i -> do
+                      b <- f i
+                      if b
+                        then pure $ error $ unwords $
+                                 ["no string value for bool field:",name]
+                        else empty
 
 
 --------------------------------------------------------------------------------


bdesham:
Any news on this PR?

bergus:
Just curious, what are you using this for?
Also what would you think of a `maybeField :: String -> (Item a -> Compiler (Maybe String)) -> Context a`

Kritzefitz:
> Just curious, what are you using this for?

I'm building a multi-level navigation menu by navigating over `Item`s. Links in the navigation that link to the current site or to sites hierarchically above the current site should be rendered in a special way. I currently do this by giving each of the menu `Item`s a field that tests for the conditions mentioned above by comparing parts of the `Item`'s `Identifier` with the current page's `Identifier`. The latter is obtained with getUnderlying, which is what I need the `Compiler` for.

> Also what would you think of a `maybeField :: String -> (Item a -> Compiler (Maybe String)) -> Context a`

Definitely looks like something that might be useful from time to time.

jaspervdj:
I'm fine merging this unless this conflicts with the progress on better error messages @bergus

bergus:
No, doesn't really conflict with that, however I'm not sold on the name `boolFieldM`. We should get a clear naming convention for how to construct `Context`s with functions that return strings, bools, lists (or `Compiler`s for them) from items, from items+arguments or from nothing.

Also I'm currently experimenting with
```hs
class ContextFieldable a where -- a horrible name. Suggestions?
    toContextField :: a -> Compiler ContextField

instance ContextFieldable ContextField where
    toContextField = return

instance ContextFieldable a => ContextFieldable (Compiler a) where
    toContextField = (>>= toContextField)

instance ContextFieldable [Char] where
    toContextField = return . StringField

instance ContextFieldable Bool where
    toContextField True  = return EmptyField
    toContextField False = compilerNoResult "false"

instance ContextFieldable a => ContextFieldable (Maybe a) where
    toContextField = maybe (compilerNoResult "false") toContextField

functionContext :: ContextFieldable c => String -> ([String] -> Item a -> c) -> Context a
functionContext key value = Context $ \k args item ->
    if k == key
        then mapError (("In field "++key):) $ toContextField $ value args item
        else compilerNoResult $ "Tried field " ++ key

context :: ContextFieldable c => String -> (Item a -> c) -> Context a
context key = functionContext key . const
```

which would allow `field = context`, `boolField = context`, `constField key = context key . const`, `functionField = functionContext`, and also **`boolFieldM = context`** - it's super-flexible. Maybe too flexible? If you want to consider this, I'm gonna file a separate pull request, superseding this one. Either is largely independent from the error messages PR though.

jaspervdj:
Oh yeah something like that would work as well.  I would just call the typeclass `ToContextField`.

LaurentRDC:
Done in #1044

PR #566, opened 2017-09-02
Fix compression of calc() in CSS

KrzysiekJ:
According to Mozilla Developer Network, “The + and - operators must always be surrounded by whitespace.”.

### DIFF
--- a/lib/Hakyll/Web/CompressCss.hs
+++ b/lib/Hakyll/Web/CompressCss.hs
@@ -1,6 +1,7 @@
 --------------------------------------------------------------------------------
 -- | Module used for CSS compression. The compression is currently in a simple
 -- state, but would typically reduce the number of bytes by about 25%.
+{-# LANGUAGE PatternGuards #-}
 module Hakyll.Web.CompressCss
     ( compressCssCompiler
     , compressCss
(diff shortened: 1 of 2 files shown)


lib/Hakyll/Web/CompressCss.hs
```diff
@@ -43,6 +45,21 @@ compressSeparators str
     stripSecond = or $ map (isOfPrefix str) $ map (\c -> c ++ " ") separators
     separators  = [" ", "{", "}", ":", ";", ",", ">", "+", "!"]
 
+-- | Compresses separators when starting inside calc().
+compressCalcSeparators :: Int -> String -> String
+compressCalcSeparators 0 str = compressSeparators str
+compressCalcSeparators depth str
+  | stripFirst = compressCalcSeparators depth (tail str)
+  | stripSecond = compressCalcSeparators depth (head str : (drop 2 str))
+  | '(' == head str = '(' : compressCalcSeparators (depth + 1) (tail str)
```
jaspervdj: Could use something like:

```
   | ('(' : xs) <- str = '(' : compressCalcSeparators (depth + 1) xs 
```

to avoid the use of partial functions?  This also applies to the `')' == head str` line.

jaspervdj:
Thanks!  I added a quick comment.

lib/Hakyll/Web/CompressCss.hs
KrzysiekJ: Ok, uploading a corrected version. Had to add `{-# LANGUAGE PatternGuards #-}` to suppress warnings.

lib/Hakyll/Web/CompressCss.hs
```diff
@@ -43,6 +45,21 @@ compressSeparators str
...
+  | isPrefixOf "calc( " str = compressCalcSeparators depth ("calc(" ++ (drop 6 str))
+  | isPrefixOf "calc(" str = '(' : compressCalcSeparators (depth + 1) (drop 5 str)
+  | ')' == head str = ')' : compressCalcSeparators (depth - 1) (tail str)
+  | otherwise = head str : compressCalcSeparators depth (tail str)
```
KrzysiekJ: This will lead to an exception if brackets do not match, like in `calc((2px)`. Is it a problem?

PR #569, opened 2017-10-07
Remove non-existent links

mgsloan:
I clicked all the links, looking for website ideas / potential code to use, removing the ones that are gone / different content. In one case, http://blog.darkfox.id.au/ , hakyll isn't being used.  In several cases, the source still exists but the website does not.

Also, when I was searching earlier I found this nice looking site that is not listed https://thibaud.dauce.fr/


jaspervdj:
Thanks for doing this!

PR #582, opened 2017-11-16
Make unixFilter output stderr when failing

### LINKED ISSUE #581: unixFilter doesn't return stderr output on failure
Hi
This is something that comes up quite a lot for us...

### Steps
1. Use unixFilter, e.g.  as per docs
```
match "style.scss" $ do
    route   $ setExtension "css"
    compile $ getResourceString >>=
        withItemBody (unixFilter "sass" ["-s", "--scss"]) >>=
        return . fmap compressCss
```
2. Include a SASS error in your `style.scss` (e.g. a line of  `@@`).

### Expected
The error message from SASS, to help us fix the problem. 

### Actual
```  [ERROR] Hakyll.Core.UnixFilter.unixFilterWith: sass -s --scss  gave exit code 65.```
You can then run SASS locally (if you're building locally - we build it in Docker so this isn't so easy) to find the output:

```
Error: Invalid CSS after "@": expected identifier, was "@@ "
        on line 1 of standard input
  Use --trace for backtrace.
```

It would be great if Hakyll (a great library - thanks BTW) passed on this output

declension:
Fixes #581.

### DIFF
--- a/lib/Hakyll/Core/UnixFilter.hs
+++ b/lib/Hakyll/Core/UnixFilter.hs
@@ -92,7 +92,8 @@ unixFilterWith writer reader programName args input = do
         ExitSuccess   -> return output
         ExitFailure e -> fail $
             "Hakyll.Core.UnixFilter.unixFilterWith: " ++
-            unwords (programName : args) ++ " gave exit code " ++ show e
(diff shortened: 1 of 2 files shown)


jaspervdj:
Thanks!

PR #584, opened 2017-11-18
Second pass at Pandoc 2.0

vaibhavsagar:
I updated `stack.yaml` to support #557.


lib/Hakyll/Web/Pandoc.hs
```diff
@@ -84,7 +86,10 @@ writePandoc = writePandocWith defaultHakyllWriterOptions
 writePandocWith :: WriterOptions  -- ^ Writer options for pandoc
                 -> Item Pandoc    -- ^ Document to write
                 -> Item String    -- ^ Resulting HTML
-writePandocWith wopt = fmap $ writeHtmlString wopt
+writePandocWith wopt (Item itemi doc) =
+    case runPure $ writeHtml5String wopt doc of
+        Left (PandocSomeError err)  -> Item itemi $ "Hakyll.Web.Pandoc.writePandocWith: unknown error: " ++ err
```
jaspervdj: We (unfortunately) want to use `error` here.  It's not a great thing to throw errors from pure code, but it's better than breaking every Hakyll site out there.

stack.yaml
```diff
@@ -2,9 +2,17 @@ resolver: nightly-2017-06-19
...
 - 'QuickCheck-2.10.0.1'
 - 'tasty-quickcheck-0.9'
+- 'pandoc-2.0.2'
+- 'cmark-gfm-0.1.3'
+- 'doctemplates-0.2.1'
+- 'hslua-0.9.2'
+- 'skylighting-0.4.3.2'
+- 'texmath-0.10'
+- 'pandoc-types-1.17.3'
+- 'hs-bibutils-6.2.0.1'
```
jaspervdj: If you bump to the latest nightly, these should now be in the index.

jaspervdj:
Thanks!

lib/Hakyll/Web/Pandoc.hs
vaibhavsagar: Updated.

stack.yaml
vaibhavsagar: Updated to `nightly-2017-11-19`.

vaibhavsagar:
Thanks for the quick review! I've made changes as requested.

vaibhavsagar:
The CI build seems to be failing because building Pandoc is taking too long, but I tested this locally and it seemed fine.

jaspervdj:
Thanks for the work!  I might put in some more changes on top of this before I put out a new release.

vaibhavsagar:
Thanks!

Issue #473, opened 2016-09-08
Tutorial missing a step?


Istap:
I set up hakyll following this [tutorial](https://jaspervdj.be/hakyll/tutorials/github-pages-tutorial.html). Since I am a complete noob I am not sure if I oversaw something obvious, but I before I was able to run `stack exec myblog clean/build` I had to run `stack build`, which created the `myblog` executable. Without this command, both clean and build failed with "Executable named clean/build not found on path".

smorele:
Is it a question ?
The `stack build` command creates an executable. You can get more informations about stack and [workflow](https://docs.haskellstack.org/en/stable/README/#workflow) on [the official website](https://docs.haskellstack.org).

christopheryoung:
I think he's pointing out that the tutorial should advise people to run `stack build` _before_ they try to run `stack exec myblog clean`. As it's currently written, it's likely to confuse people.

mchowning:
Added a PR to address this: https://github.com/jaspervdj/hakyll/pull/588

walambert:
Hi,
The tutorial is still missing the "stack build" step, it would be quite helpful if the tutorial included that. This page (#473) and #588 show up relatively quickly on a google search, but not always, so including that step on the tutorial page would streamline the hakyll setup process for many (especially beginners like myself). Thanks :)

jaspervdj:
Done, sorry for the delay and thanks for the reminders!

PR #588, opened 2017-11-26
Add 'stack build' step to github pages tutorial

mchowning:
Addressing #473

### DIFF
--- a/web/tutorials/github-pages-tutorial.md
+++ b/web/tutorials/github-pages-tutorial.md
@@ -76,6 +76,12 @@ Ensure we are in the correct branch.
 git checkout develop
 ```
 
+If this is a new Hakyll project, run Stack's build command.
+
+```
+stack build
+```
+
 Get a clean build of our site.
 
 ```


christopheryoung:
Should the tutorial also mention that the `stack build` command will create an executable that you will need to put in the project directory for the `stack exec myblog build` command to work? (I might just be confused, but that's what I had to do.)

mchowning:
@christopheryoung : that's interesting, I didn't have to move the executable in order for `stack exec myblog clean` to work after running `stack build`, and I feel like you shouldn't have to do that, but I'm not a Stack expert.

As I understand it, `stack build` creates the executable within in the project's `.stack-work` directory and that `stack exec` can run the executable from that location (see [here](https://docs.haskellstack.org/en/stable/GUIDE/#stack-exec)). 

For example, when I run `stack exec myblog clean`on a new Hakyll project based on Stack's `hakyll-template` (without running `stack build`) I see an error message that Stack is unable to find the executable in a number of different directories, the first of which is within the project's `.stack-work` directory:

```
▶ stack exec myblog clean
Executable named myblog not found on path: 
["{myblog_project_directory}/.stack-work/install/x86_64-osx/lts-9.14/8.0.2/bin", ...]
```

Running `stack build` installs the project's executable into that first directory where the `stack exec myblog clean` command was looking for the executable:

```
▶ stack build
...
Installing executable(s) in
{myblog_project_directory}/.stack-work/install/x86_64-osx/lts-9.14/8.0.2/bin
```

Therefore, after running `stack build`, `stack exec myblog clean` runs successfully. 

So I think that `stack build` should be sufficient on its own to create the executable at a location that is accessible by `stack exec`, but let me know if I'm missing something.

christopheryoung:
@mchowning, ah, I see. I think I was confused by some out of date documentation that talked about copying over the executable. Thank you for the explanation.

manuel-uberti:
Thanks for this, I was stuck as well the "executable not found" error.

PR #602, opened 2018-01-03
Load a template from a local files if present then fallback to embedded files

kuznero:
This is very useful when hakyll is used to build statically linked binary in which case it would make sense to have a way to provide template files from a current working directory instead of fetching it from `data-files` that is very specific to where haskell code was built (e.g. build server).

I have elaborated on this a bit in my blog post [here](http://www.kuznero.com/posts/haskell/building-statically-linked-binaries.html#data-files-in-dependencies).

### DIFF
--- a/hakyll.cabal
+++ b/hakyll.cabal
@@ -56,10 +56,6 @@ Data-files:
   example/index.html
   example/about.rst
   example/contact.markdown
-  templates/atom-item.xml
-  templates/atom.xml
-  templates/rss-item.xml
-  templates/rss.xml
 
 Extra-source-files:
   CHANGELOG.md
@@ -190,7 +186,8 @@ Library
     unordered-containers >= 0.2      && < 0.3,
     vector               >= 0.11     && < 0.13,
     yaml                 >= 0.8.11   && < 0.9,
-    optparse-applicative >= 0.12     && < 0.15
+    optparse-applicative >= 0.12     && < 0.15,
+    file-embed           >= 0.0.10.1 && < 0.0.11
 
   If flag(previewServer)
     Build-depends:
--- a/hakyll.cabal
+++ b/hakyll.cabal
@@ -73,6 +73,10 @@ Extra-source-files:
   tests/data/strip.html.out
   tests/data/template.html
   tests/data/template.html.out
+  data/templates/atom-item.xml
+  data/templates/atom.xml
+  data/templates/rss-item.xml
+  data/templates/rss.xml
 
 Source-Repository head
   Type:     git
--- a/lib/Hakyll/Web/Template/Internal.hs
+++ b/lib/Hakyll/Web/Template/Internal.hs
@@ -11,6 +11,7 @@ module Hakyll.Web.Template.Internal
     , applyAsTemplate
     , readTemplate
     , unsafeReadTemplateFile
+    , unsafeReadTemplateFile'
 
     , module Hakyll.Web.Template.Internal.Element
     , module Hakyll.Web.Template.Internal.Trim
@@ -24,6 +25,7 @@ import           Data.List                            (intercalate)
 import           Data.Typeable                        (Typeable)
 import           GHC.Exts                             (IsString (..))
 import           Prelude                              hiding (id)
+import           System.Directory                     (doesFileExist)
 
 
 --------------------------------------------------------------------------------
@@ -201,3 +203,12 @@ unsafeReadTemplateFile file = do
     tpl <- unsafeCompiler $ readFile file
     pure $ template $ readTemplateElemsFile file tpl
 
+
+--------------------------------------------------------------------------------
+unsafeReadTemplateFile' :: FilePath -> String -> Compiler Template
+unsafeReadTemplateFile' file defValue = do
+    tpl <- unsafeCompiler $ do
+      isLocal <- doesFileExist file
+      if isLocal then readFile file
+      else return defValue
+    pure $ template $ readTemplateElemsFile file tpl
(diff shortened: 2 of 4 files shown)


jaspervdj:
Hmm, I think it would be even nicer if we could use [file-embed](https://hackage.haskell.org/package/file-embed) to pull in the data-files.  That way we don't really need to worry about them being there at all.  What do you think?

kuznero:
Using `file-embed` instead of relying on `data-files` together with checking local files in order to make it possible to modify default templates would make sense absolutely. There is only one _"but"_, `file-embed` is yet another dependency. Not that it is problematic, though it is using _TemplateHaskell_ which will make the build time slightly longer.

But if that is not an issue, I can prepare the version with `file-embed`, no problem.

jaspervdj:
We already depend on `file-embed` indirectly (through `wai-app-static` and maybe other paths) so I think it's fine.  I think the extra compile time is worth the fact that people can just `scp` their executable without worrying too much about the Hakyll templates being installed in the right location.

kuznero:
Will do that shortly. As I am developing using NixOS, I introduced some files with nix-expressions to help develop under NixOS. Do you mind if it will be added as part of this PR or I can simply strip it away completely?

kuznero:
Done. @jaspervdj, please review.

lib/Hakyll/Web/Template/Internal.hs
```diff
@@ -201,3 +203,12 @@ unsafeReadTemplateFile file = do
...
 
+
+--------------------------------------------------------------------------------
+unsafeReadTemplateFile' :: FilePath -> String -> Compiler Template
+unsafeReadTemplateFile' file defValue = do
+    tpl <- unsafeCompiler $ do
+      isLocal <- doesFileExist file
+      if isLocal then readFile file
+      else return defValue
+    pure $ template $ readTemplateElemsFile file tpl
```
jaspervdj: I would put this function in the `where` clause of `renderFeed`, since we're not really using it anywhere else, and the name isn't really general enough to be true general-purpose.

hakyll.cabal
```diff
@@ -56,10 +56,6 @@ Data-files:
   example/index.html
   example/about.rst
   example/contact.markdown
-  templates/atom-item.xml
-  templates/atom.xml
-  templates/rss-item.xml
-  templates/rss.xml
```
jaspervdj: These files need to be added to `Extra-source-files`, otherwise they don't can included in the tarball when running `cabal sdist`.

kuznero:
As a result there are some additional imports. Please review.

jaspervdj:
Hey @kuznero, thanks for the patches!

After thinking about it for a bit, I'm not sure if we want to allow changing the default templates this way.  It doesn't sound very haskell-y to create a file with a special name in the filesystem and suddenly everything looks different!

If we really want people to be able to customize the templates (which I'm not so sure about, atom and RSS are pretty rigid), this should be parameters to `renderRss` and `renderAtom`.  What do you think?  Do you have a concrete use case for wanting to customize the templates?

kuznero:
I started changing it in the first place because I needed to change item formats:

In `rss-item.xml` with `description` having `$body$` instead of `$description$`:

```xml
<item>
    <title>$title$</title>
    <link>$root$$url$</link>
    <description><![CDATA[$body$]]></description>
    <pubDate>$published$</pubDate>
    <guid>$root$$url$</guid>
    <dc:creator>$authorName$</dc:creator>
</item>
```

And also in `atom-item.xml` with `summary` having `$body$` instead of `$description$`:

```xml
<entry>
    <title>$title$</title>
    <link href="$root$$url$" />
    <id>$root$$url$</id>
    <published>$published$</published>
    <updated>$updated$</updated>
    <summary type="html"><![CDATA[$body$]]></summary>
</entry>
```

So, going from `data-files` to simply embedding templates was right. Question is whether to let others change files or instead extend haskell API. And frankly, I agree with you, I think the right approach is to extend haskell API especially for such rigid format as RSS and ATOM.

Another question though - can we also change the default from `$description$` to `$body$`?

jaspervdj:
@kuznero You can do that without changing Hakyll.  I am also doing that for my personal site, see the code here:

1. I render the RSS feed with a special `feedCtx`: <https://github.com/jaspervdj/jaspervdj/blob/master/src/Main.hs#L150>
2. In `feedCtx`, I put the body in `$description$`: <https://github.com/jaspervdj/jaspervdj/blob/master/src/Main.hs#L221-L225>

Doing that should solve your problem!

I still want to merge this PR, just having `embed-file` is quite useful I think since then you get an executable which doesn't depend on any other files.  Do you want to update the PR to remove the customization/fallback?

kuznero:
@jaspervdj thanks for recommendation with regards to `feedCtx`. Otherwise, I updated PR to have only embedded templates for the feeds.

kuznero:
I have tried using `feedCtx` defining `body` instead of `description` and it does not seem to work for me:

```haskell
let feedCtx = bodyField "body" <> defaultContext <> postCtx
posts <- fmap (take 100) . recentFirst =<< loadAll pat
renderRss feedcfg feedCtx posts
```

Am I doing something wrong? `rss.xml` generated still have short `description` fields instead of `body` that I want.

jaspervdj:
Well, what you want to do is put `bodyField "description"` first, so when it the template looks up `$description$`, it hits the body.  E.g.:

```haskell
let feedCtx = bodyField "description" <> defaultContext <> postCtx
posts <- fmap (take 100) . recentFirst =<< loadAll pat
renderRss feedcfg feedCtx posts
```

kuznero:
@jaspervdj that worked, thanks!

PR #611, opened 2018-02-11
Add Semigroup instances for existing Monoids

### LINKED ISSUE #525: Add Semigroup instances

cbarcenas:
Ensures forwards compatibility with future Haskell/GHC releases as the Semigroup/Monoid Proposal is gradually implemented.

Closes #525 and #536.

### DIFF
--- a/lib/Hakyll/Check.hs
+++ b/lib/Hakyll/Check.hs
@@ -19,6 +19,9 @@ import           Control.Monad.Trans          (liftIO)
 import           Control.Monad.Trans.Resource (runResourceT)
 import           Data.List                    (isPrefixOf)
 import qualified Data.Map.Lazy                as Map
+#if MIN_VERSION_base(4,9,0)
+import           Data.Semigroup               (Semigroup (..))
+#endif
 import           Network.URI                  (unEscapeString)
 import           System.Directory             (doesDirectoryExist,
                                                doesFileExist)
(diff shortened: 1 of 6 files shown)


bergus:
Can you avoid the code duplication by using the [recommended pattern](https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid#RecommendedVariant) from the proposal?
```hs
instance Semigroup Foo where
  (<>) = …

instance Monoid Foo where
  mempty = …

#if !(MIN_VERSION_base(4,11,0))
  mappend = (<>)
#endif
```

jaspervdj:
The build is failing on `lib/Hakyll/Core/Identifier/Pattern.hs` because there is a `/*` character sequence in a _Haskell_ comment, which CPP doesn't understand so it bails out.

There's not much we can do about this aside from cursing at CPP and moving the "guts" of the module (`Pattern` and its instances) to `lib/Hakyll/Core/Identifier/Pattern/Internal.hs` which is allowed to have CPP, and then re-export it from `lib/Hakyll/Core/Identifier/Pattern.hs` where the comment lives.

Would you be willing to update the PR to that?

ChShersh:
You can also configure `cpphs` to be C-preprocessor inside `.cabal` file. `cpphs` can handle `/*` properly (most likely).

PR #572, opened 2017-10-15
Add nested metadata parsing

ilya-murzinov:
Resolves #528

Issue #529, opened 2017-03-10
allow supplying a custom metadata format parser for compiler


lennart:
currently hakyll seems to be only able to read metadata as YAML (as far as I understand the code in `Provider/Metadata.hs`)

I want to be able to have hakyll keep source code function documentation in metadata as key value pairs (functionname: functiondocstring) when compiling source code files.

big picture is 
- [x] route to a pattern of haskell source files, compile with a custom `Compiler`
- [ ] then build a map of all functions (docs) found in all files given by a pattern (similar to `buildTags` but not quite)
- [x] then write function references within markdown pages as standard yaml metadata list field
- [x] build a map of those function references
- [ ] match them against the function docs by their identifiers (function names)
- [ ] render the markdown page with content & a list of rendered functionname/functiondocstrings

So in order to do that (as the docstrings do not appear in single files and I cannot find another way to do the above) I was thinking of a custom Metadata parser (I can provide) that essentially parses haskell source code into an AST and extracts the function names and docstrings into a `Metadata` object.

How would that be possible? I am not hesitant to put this into a contrib module, as I fear this is a rather non-standard feature request.

However right now, I think I would need to fork hakyll in order to do that, and I would rather not.

suggestions on how to proceed?

kindly,

lennart

jaspervdj:
I don't think you necessarily need to do this through metadata, I think you can approach it from a different angle as well.

What I suggest is the following flow:

1. Compile all source files to a custom datatype (e.g. `DocMap`).  Use a `version "docmap"` here to distinguish it.
2. Add a compiler which loads all `DocMap`s produced in (1) and `mconcat`s them together to get a "global" `DocMap`.
3. Compile all source files "normally" but load the global `DocMap` so you can use that to render the table of context and so on.

PR #620, opened 2018-03-17
Fancy generic templating stuff for nested data

bergus:
This takes various ideas from #529, #572 and #565 into action.

I've got a data file (json, or better yaml as that has comments) in my project whose contents I want to use in rendering various files - both html and json. Call it "shared (meta)data" if you want.

My first idea was a classic `Compiler` - parse the JSON file using Aeson, store in a snapshot, spit back out (without whitespace). Access the parsed data from the snapshot elsewhere (in the html templates).
Problem: snapshotted data must be a `Binary` instance. Hakyll can do this already with data, but the `BinaryYaml` (a newtype wrapper over `Value`) is not exported.

My second idea was using Hakyll's builtin metadata parsing. Have a `xyz.metadata` file, use `getMetadata "xyz"` in the Compiler monad. Hakyll would do the parsing and caching for me, right?
Problem: `xyz` does not exist as a resource, so no metadata can be read from it (instead it always comes back as an empty object by default).

I guess this could have been solved with an empty `xyz` file, but instead I went ahead and <s>fixed</s> updated the `Provider` to consider it as existing when there's a `xyz.metadata` file.

Next step: write a custom context that allows me to access *and iterate* deep (nested) `Value`s - generically, without parsing the JSON in a dedicated structure and supplying the relevant list fields.
Converting Aeson `Value`s into the respective strings was cumbersome though, and Hakyll already does have `toString` which I didn't want to replicate. Also this functionality seemed so useful that it might be interesting to add natively, possibly also for `Metadata` like #572.
I've got a proof of concept working in the attached commits (starting at "WIP", the rest is from #462), but this is much less a pull request than a **request for comments** as of now. What do you all think? Specifically

* do we need to support keys that contains dots? If yes, how - and does it need to be backwards-comptible?
* what fields should be available inside a `$for(…)$` loop that iterates an array, or an object? Should the outer context be sustained, i.e. if `$var$` is available outside of the loop should it be accessible from inside the loop as well?


bergus:
@jaspervdj What do you think about 73cc078? This seemed like the right idea, but my understand of `Provider` is too shallow yet. Specifically, treating identifiers as resources just because a `.metadata` file exists for them might have some weird side effects - no file at that path exists, but `resourceExists` says otherwise. Also it does get matched by wildcard patterns that do not contain a file extension, and trying to compile such a non-existing file will fail.
Is my commit an appropriate hack, and the errors it could cause are warranted?
Or does this need a more holistic design approach - I also noticed there's still a `TODO` in the code about modification dates of metadata (in general, not only external files). Should metadata and contents (bodies) be versioned separately, as they are already cached separately?

jaspervdj:
I like what this PR tries to accomplish -- it's something I've also wanted for a long time.

- do we need to support keys that contains dots? If yes, how - and does it need to be backwards-comptible?

No, I don't think we need to support this.  In the worst case we can add escaping `"foo\.bar" later if it turns out to be truly necessary.

- what fields should be available inside a $for(…)$ loop that iterates an array, or an object? Should the outer context be sustained, i.e. if $var$ is available outside of the loop should it be accessible from inside the loop as well?

I think it makes sense to do lexical scoping -- so the outer context should be sustained, but the inner context can overwrite it if there's a conflict.

Going to have a look at the code now.

lib/Hakyll/Core/Provider/Internal.hs
```diff
@@ -120,20 +118,20 @@ newProvider store ignore directory = do
...
 --------------------------------------------------------------------------------
-getResourceInfo :: FilePath -> Set Identifier -> Identifier -> IO ResourceInfo
-getResourceInfo directory universe identifier = do
-    mtime <- fileModificationTime $ directory </> toFilePath identifier
-    return $ ResourceInfo (BinaryTime mtime) $
-        if mdRsc `S.member` universe then Just mdRsc else Nothing
-  where
-    mdRsc = fromFilePath $ flip addExtension "metadata" $ toFilePath identifier
+getResourceInfo :: FilePath -> Identifier -> IO (Identifier, ResourceInfo)
+getResourceInfo directory identifier = do
```
jaspervdj: Do we need to support asking for resource info about `.metadata` files?

lib/Hakyll/Web/Feed.hs
```diff
@@ -41,20 +42,20 @@ import qualified Data.Text.Encoding          as T
 
 
 --------------------------------------------------------------------------------
-rssTemplate :: String
-rssTemplate = T.unpack $
+rssTemplate :: Item String
+rssTemplate = Item "templates/rss.xml" $ T.unpack $
```
jaspervdj: Maybe we should call these something like `Item "_hakyll/templates/rss.xml"` so they can't conflict with actual user-provided items?

lib/Hakyll/Web/Template/Context.hs
```diff
@@ -78,71 +112,133 @@ newtype Context a = Context
...
 --------------------------------------------------------------------------------
-functionField :: String -> ([String] -> Item a -> Compiler String) -> Context a
-functionField name value = Context $ \k args i ->
-    if k == name
-        then StringField <$> value args i
-        else empty
+-- | Creates a variadic function field.
+--
+-- The function will be called with the dynamically evaluated string arguments
+-- from the template as well as the page that is currently rendered.
```
jaspervdj: Thanks for improving the docs!

lib/Hakyll/Web/Template/Internal.hs
```diff
@@ -1,15 +1,20 @@
...
 {-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE DeriveAnyClass             #-}
+{-# LANGUAGE DeriveGeneric              #-}
+
 module Hakyll.Web.Template.Internal
-    ( Template (..)
+    ( Template
     , template
+    , unTemplate
+    , getOrigin
```
jaspervdj: Do we need to export `unTemplate`?  In either case I'd call these `templateElements` and `templateOrigin`.

lib/Hakyll/Web/Template/Context.hs
```diff
@@ -151,15 +247,67 @@ mapContext f (Context c) = Context $ \k a i -> do
...
+
+
+
+dataField :: String -> Value -> Context a
+dataField key val = Context $ \f a (Item i _) -> case splitAll "." f of
+    [k] | k == get    -> lookupNestedValue a  (Item i val)
+    (k:ks) | k == key -> lookupNestedValue ks (Item i val)
+    _                 -> failBranch $ "Tried field " ++ key -- and functionField get
+  where
+    get = let (h:rest) = key in "get" ++ toUpper h : rest
```
jaspervdj: I don't really understand what this does :thinking: 

Are we encoding `getFoo` as special keys?

jaspervdj:
I think the cleanest thing would probably to consider a `.metadata` file and the resource as "one thing".

lib/Hakyll/Web/Feed.hs
bergus: Ah, this change is actually part of #462 on which this PR is based on :-)
I think the `Item` here is solely used for generating template error messages, it never faces the user anywhere. Or am I missing something that `Compiler` internals would do to an `Item`?

lib/Hakyll/Web/Template/Internal.hs
bergus: I think we need to export them from `Template.Internal`, yes. About renaming them: `unTemplate` had been used and exported before as part of the non-smart `Template` constructor.

lib/Hakyll/Web/Template/Context.hs
bergus: This allows one to create a `dataField "foo" (Object metadata)` or something, then use `$foo.bar.baz$` in templates or `$getFoo("bar", "baz")$` - this allows lookup with dynamic names (generated by another field) and also supports names with dots (or any other special syntax).

lib/Hakyll/Core/Provider/Internal.hs
bergus: I guess I should rename this function to `getFileInfo`. It is called on every (non-ignored) file in the directory, and then it decides itself whether that file is metadata of another resource or a resource itself.

bergus:
> I think the cleanest thing would probably to consider a .metadata file and the resource as "one thing".

Yes, that's what I've been trying to do with `M.fromListWith combine <$> mapM (getResourceInfo directory) list` and the `combine` function.
The problem is that I don't know which code in Hakyll does depend on the assumption that every "thing" (`Identifier`) has a resource.

bergus:
> I think it makes sense to do lexical scoping

Yes, I figured that out when trying it - I need to have some control over the contexts in the loops.

That means my current approach with `Context Value`, `Context (Int, Value)` and `Context (Text, Value)` won't work, I need to pass the value to the creation of the context itself not extract it from the Item.

lib/Hakyll/Web/Template/Context.hs
```diff
@@ -92,7 +94,7 @@ data ContextField
     = EmptyField
     | StringField String
     | forall a. ListField (Context a) [Item a]
-
+    | forall a. LexicalListField (forall b. Context b -> a -> Context b) [a]
```
bergus: Lexical scoping was not possible with the current `ListField` constructor, needed something more advanced :-) Also requires RankNTypes, not sure whether that's a bad thing.
Maybe it's also possible to further generalise this, but I don't see a way to do it without breaking compatibility for everyone who directly uses `ListField`. Maybe these should be moved to a `Context.Internal` module anyway.

lib/Hakyll/Web/Template/Context.hs
```diff
@@ -21,9 +21,13 @@
 --------------------------------------------------------------------------------
 {-# LANGUAGE CPP                       #-}
 {-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleInstances #-}
 module Hakyll.Web.Template.Context
     ( ContextField (..)
     , Context (..)
+    , context
+    , functionContext
```
bergus: I wonder whether we should export the general functions under new names. I don't really like `context` and `functionContext`, as they construct contexts for specific keys. Should be `keyContext` at best.
Do you think it would do any harm to just widen the type of `field` and `functionField`?

jaspervdj: I think it's fine to widen the type if it doesn't break existing code.  I'm a bit afraid it could break existing code though, since people usually use `{-# LANGUAGE OverloadedStrings #-}`.

nicola-gigante:
Is there any news on this functionality?

PR #622, opened 2018-03-19
make pandoc optional

bergus:
Cuts down my build size from 41,005,585 Bytes to 31,203,857 Bytes.

### DIFF
--- a/hakyll.cabal
+++ b/hakyll.cabal
@@ -174,12 +174,10 @@ Library
     filepath             >= 1.0      && < 1.5,
     lrucache             >= 1.1.1    && < 1.3,
     mtl                  >= 1        && < 2.3,
-    network              >= 2.6      && < 2.7,
     network-uri          >= 2.6      && < 2.7,
     parsec               >= 3.0      && < 3.2,
     process              >= 1.6      && < 1.7,
     random               >= 1.0      && < 1.2,
-    regex-base           >= 0.93     && < 0.94,
     regex-tdfa           >= 1.1      && < 1.3,
     resourcet            >= 1.1      && < 1.3,
     scientific           >= 0.3.4    && < 0.4,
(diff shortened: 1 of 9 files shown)


hakyll.cabal
```diff
@@ -99,6 +99,10 @@ Flag buildWebsite
   Default:     False
   Manual:      True
 
+Flag usePandoc
+  Description: Include Pandoc support
+  Default:     True
```
jaspervdj: We should also make this a `Manual: True` flag

tests/Hakyll/Web/Template/Tests.hs
```diff
@@ -89,7 +88,9 @@ test (outf, tplf, itemf) = do
     out  <- resourceString provider outf
     tpl  <- testCompilerDone store provider tplf templateBodyCompiler
     item <- testCompilerDone store provider itemf $
-        pandocCompiler >>= applyTemplate (itemBody tpl) testContext
+        getResourceBody
+        >>= renderParagraphs
```
jaspervdj: I would put a comment here about why we're not using pandoc

bergus:
OK, updated.

jaspervdj:
Thanks again!

bergus:
What is weird though is that I'm getting a warning
```
$ stack install
hakyll-4.12.0.1: build (lib + exe)

Warning: The following modules should be added to exposed-modules or other-modules in C:\Users\…\hakyll\hakyll.cabal:
             - In hakyll-tests component:
                 Hakyll.Web.Pandoc.FileType.Tests

         Missing modules in the cabal file are likely to cause undefined reference errors from the linker, along with other problems.
hakyll-4.12.0.1: copy/register
```
when I run my project with `usePandoc: False`

jaspervdj:
I guess `Hakyll.Web.Pandoc.FileType` doesn't _really_ depend on pandoc.  I'll fix this.

PR #634, opened 2018-04-28
Fix path resolution in Pandoc.Biblio

### LINKED ISSUE #195: Pandoc.Biblio doesn't work with non-standard provider directory
If `providerDirectory` is set to a nonstandard location (e.g. `content`), the `Pandoc.Biblio` code fails with file-not-found errors loading the CSL and the BibTeX.

It seems that it assumes the file path is relative to the current directory, not the provider directory.

knih:
Fixes #195


jaspervdj:
Thank you!

Issue #225, opened 2014-02-23
YAML metadata?


thejohnfreeman:
The metadata supported is not complete YAML. It does not handle lists spanning multiple lines:

```
entries:
- one
- two
- three
```

The error:

```
unexpected "\n"
expecting space followed by metadata for: entries
```

I got the feeling the metadata parser is something custom written for Hakyll. Would it be possible to just use a general YAML parser?

In my particular case, I have about 7 entries, each with multiple fields, that I use to conveniently generate a portion of my page. It's going to look ugly and unmaintainable if consolidated to a single line.

FranklinChen:
I just ran across this incompletely YAML support also, when attempting to start migrating an Octopress-based site to Hakyll.

jmatsushita:
+1 for this. Trying to migrate from metalsmith with YAML frontmatters as well!

gpampara:
+1 Just hit this issue as well. Would be great to get this corrected

vjeranc:
Would fixing this enable easier references (without bib and cls files) using YAML metadata as described [here](http://pandoc.org/README.html#citations)?

hazybluedot:
+1 again, just ran into this trying to port my blog over from nanoc!

justinmanley:
+1 Full YAML front-matter support would be great!

jaspervdj:
Hakyll 0.8.0.0 turns `Metadata` into a YAML object.

thejohnfreeman:
Perfect timing! Just last week I returned to Hakyll after more studiously learning Haskell over the previous months, with the intention of addressing this issue.

beerendlauwers:
Stellar. This is a long-standing issue that made Hakyll a hard sell to some clients and managers.

PR #637, opened 2018-05-29
Update tutorial for YAML object metadata

fschlaepfer:
Since version 4.8 `Metadata` is no longer a Map but a YAML object. (see #225)

The compilers tutorial is adapted to use `lookupString` instead of `Map.lookup`.


jaspervdj:
Thanks!

PR #641, opened 2018-06-09
Use makeRelativeToProject with embedFile (for stack ghci)

mgsloan:
The 'makeRelativeToProject' allows usage of package relative filepaths, even if
ghc's working dir is not the package directory.  This enables me to have a
locally modified version of hakyll as part of my stack project, and run "stack
ghci" to load my hakyll project + hakyll all into one ghci session.


jaspervdj:
Thanks!

PR #645, opened 2018-06-28
Fix paths for compatibility with Windows

### LINKED ISSUE #644: The latest Hakyll.Web.Feed can not be compiled on Windows.
**Where there is a problem:**

https://github.com/jaspervdj/hakyll/blob/1abdeee743d65d96c6f469213ca6e7ea823340a7/lib/Hakyll/Web/Feed.hs#L44-L58

**Problem:**

```
    C:\(private info)\hakyll\lib\Hakyll\Web\Feed.hs:46:18: error:
        ? Exception when trying to run compile-time code:
            C:\(private info)\hakyll\data/templates/rss.xml: openBinaryFile: does not exist (No such file or directory)
          Code: makeRelativeToProject "data/templates/rss.xml" >>= embedFile
        ? In the untyped splice:
            $(makeRelativeToProject "data/templates/rss.xml" >>= embedFile)
       |
    46 |     T.decodeUtf8 $(makeRelativeToProject "data/templates/rss.xml" >>= embedFile)
       |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This line causes this kind of error in Windows. I want to do a workaround: change the path with CPP. If you do not mind, I will send a pull request.

Hexirp:
Fix #644.

### DIFF
--- a/lib/Hakyll/Web/Feed.hs
+++ b/lib/Hakyll/Web/Feed.hs
@@ -38,46 +38,25 @@ import           Hakyll.Web.Template.List
 import           Data.FileEmbed              (makeRelativeToProject, embedFile)
 import qualified Data.Text                   as T
 import qualified Data.Text.Encoding          as T
+import           System.FilePath             ((</>))
 
 
 --------------------------------------------------------------------------------
-#if !defined(mingw32_HOST_OS)
-
 rssTemplate :: String
 rssTemplate = T.unpack $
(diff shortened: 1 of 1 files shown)


jaspervdj:
@Hexirp Thanks for the PR! I think we can actually do this in a nicer way using `</>` or `joinPath` from `System.FilePath`.  That way we can avoid the CPP requirement.

Hexirp:
Thanks for suggestion! I updated.

PR #646, opened 2018-07-01
Use Data.ByteString.Base16.encode.

0xd34df00d:
I have a somewhat pages lookup-heavy site (lots of structure with autogenerated TOCs), and profiling shows `hash` is responsible for 40% of MUT time and roughly 45% of allocations. Replacing `concatMap` over `printf` with a bytestring-operating base16 encoding makes `hash` go away from the first lines of the profile and reduces site build time from ~10 seconds to 5 seconds for me (which looks like a nice improvement!).

I guess the names of the cached data are changed with this modification, but that's a matter of `site clean` (or `site rebuild`) anyway.

Hopefully you don't mind an additional dependency on base16-bytestring!

### DIFF
--- a/lib/Hakyll/Core/Store.hs
+++ b/lib/Hakyll/Core/Store.hs
@@ -27,6 +27,7 @@ import           Data.Maybe           (isJust)
 import qualified Data.Text            as T
 import qualified Data.Text.Encoding   as T
 import           Data.Typeable        (TypeRep, Typeable, cast, typeOf)
+import           Numeric              (showHex)
 import           System.Directory     (createDirectoryIfMissing)
 import           System.Directory     (doesFileExist, removeFile)
 import           System.FilePath      ((</>))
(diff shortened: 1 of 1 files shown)


jaspervdj:
@0xd34df00d Thanks a lot for looking into this!  I had no idea this was a problem.  Can you try the following?

```haskell
import Numeric (showHex)

hash = toHex . ...
  where
    toHex :: [Word8] -> String
    toHex [] = ""
    toHex (x : xs)
        | x < 16    = '0' : showHex x (toHex xs)
        | otherwise = showHex x (toHex xs)
```

I _think_ that should remove a lot of allocations without requiring everyone to rebuild their site.  If it doesn't help, I'm happy merging in your patch.

0xd34df00d:
Hey @jaspervdj , sorry for taking so long to try out your suggestion! Indeed, it's improving things too quite a lot! The approach in my patch is just marginally faster (~0.5s on my site), so I'd be happy to go with your suggestion!

Would you do your change to upstream hakyll, or shall I issue another PR/update this one?

jaspervdj:
@0xd34df00d Can you update this PR or close it and create a new one (whichever is easier for you)?  That way you get proper credit.  :-)

0xd34df00d:
@jaspervdj sure! Done!

jaspervdj:
Thanks!

PR #652, opened 2018-07-28
Adds support for rendering feeds with custom templates

### LINKED ISSUE #625: Add tags/categories to feeds
While migrating from ikiwiki, I see that hakyll's feeds are not including tag information. This is how ikikiwi renders the categories:

```xml
<entry>
  <id>http://…</id>
  <link href="…"/>
  <category term="opensource" />
  <updated>…</updated>
  …
</entry>
```

So the only change needed would be to add either tags (my preference) or categories to the atom-item/rss-item template.

What do you think?

abhin4v:
An example usage is as follows:

```haskell
customRenderAtom :: FeedConfiguration -> Context String -> [Item String] -> Compiler (Item String)
customRenderAtom config context items = do
  atomTemplate     <- unsafeCompiler $ readFile "templates/atom.xml"
  atomItemTemplate <- unsafeCompiler $ readFile "templates/atom-item.xml"
  renderAtomWithTemplates atomTemplate atomItemTemplate config context items
```

This should also fix #625 as people will be able to add custom templates with tags.

### DIFF
--- a/lib/Hakyll/Web/Feed.hs
+++ b/lib/Hakyll/Web/Feed.hs
@@ -22,6 +22,8 @@ module Hakyll.Web.Feed
     ( FeedConfiguration (..)
     , renderRss
     , renderAtom
+    , renderRssWithTemplates
+    , renderAtomWithTemplates
     ) where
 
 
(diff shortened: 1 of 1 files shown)


abhin4v:
@jaspervdj any comments on this?

rabbitonweb:
@jaspervdj I would be happy to see this feature in. Anything blocking it? Can we help with anything to push this through?

Issue #640, opened 2018-06-04
toFilePath for Identifier returns a file path separated by slash even on Windows


kakkun61:
```haskell
Hakyll.Core.Identifier.toFilePath :: Identifier -> FilePath
```

`toFilePath` returns a file path separated by slash even on Windows.
It may be expected separated by backslash.

Is this a bug or a spec?

kakkun61:
https://github.com/jaspervdj/hakyll/pull/664 closes this issue.

PR #664, opened 2018-09-29
Fix about file path for compatibility with Windows

Hexirp:
This commit fixes two bugs related to path separator.

1. The 'makeRelativeToProject' reads files according to those filepaths. Therefore, compilation will fail if its path separator are wrong.
2. When converting to the 'Identifier', incorrect path separator are mixed, which causes file loading to fail.

See also: #640, #644, #645, #649

### DIFF
--- a/lib/Hakyll/Core/Identifier.hs
+++ b/lib/Hakyll/Core/Identifier.hs
@@ -21,7 +21,8 @@ module Hakyll.Core.Identifier
 --------------------------------------------------------------------------------
 import           Control.DeepSeq     (NFData (..))
 import           Data.List           (intercalate)
-import           System.FilePath     (dropTrailingPathSeparator, splitPath)
+import           System.FilePath     (dropTrailingPathSeparator, splitPath,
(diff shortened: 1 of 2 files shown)


jaspervdj:
Hi @Hexirp, and thanks for your help again!

I'm happy to merge in the changes wrt. the templates.  However, the idea is that `Identifier` will always have `/`-separated paths.  This probably means that the current implementation of `toFilePath` is wrong, and not `fromFilePath`?

Hexirp:
Surely. I will fix the patch accordingly. However, that change will cause duplication. I might be able to solve it with List.

Hexirp:
And, this pull request become to changes the types of `Identifier, identifierPath`. It is necessary to bump the version.

Hexirp:
A year-long reminder: @jaspervdj .

I rebased and updated this pull request. This pull request only changes `toFilePath` and `Hakyll.Web.Feed`.

jaspervdj:
Thank you!

PR #666, opened 2018-09-30
Drop extension when parsing dates in filepaths

GAumala:
Drop the file extension when parsing the date from the filepath of a resource.

This fixes a bug in which Hakyll is unable to parse the date from a filepath and throws an error when the file has a name like `posts/2018-09-26.md` because the extension becomes part of the day value.


jaspervdj:
Awesome, this looks like a perfect fix and thanks for including a test!

PR #672, opened 2018-10-13
Updated type of fromFilePath to use FilePath instead of String, resolves #661

### LINKED ISSUE #661: Type of fromFilePath
In `Hakyll.Core.Identifier`, the `fromFilePath` function has the type `String -> Identifier`. Is there any reason this isn't `FilePath -> Identifier`? As it stands I think it's confusing/misleading, and the adjacent `toFilePath` function is correctly typed as `Identifier -> FilePath`.

jhmcstanton:


jaspervdj:
Thanks!

PR #678, opened 2018-12-08
Expose `getCategory` in Haskell.Web.Tags

wei2912:
`getCategory` functions similarly to `getTags` and seems to be a pretty useful function to have.

### DIFF
--- a/lib/Hakyll/Web/Tags.hs
+++ b/lib/Hakyll/Web/Tags.hs
@@ -43,6 +43,7 @@
 module Hakyll.Web.Tags
     ( Tags (..)
     , getTags
+    , getCategory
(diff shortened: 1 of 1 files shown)


HaoZeke:
@jaspervdj  please merge this...

bergus:
Dunno, this just gets a file path from an identifier. Due to its signature (`m []`) it doesn't look particularly helpful outside of `buildCategories`. What do you need it for?

duckwork:
I know that I have a rewritten `getCategory` in my site's source because I use it to make a plain category field (i.e., not wrapped in a link). It might also be useful if someone wants to wrap the category field in a custom way.

PR #685, opened 2019-02-07
docs: Update tutorial for consistent stack use

HaoZeke:
This is more consistent since the previous instructions use `stack`. It's also a huge trip up if multiple versions are somehow installed (like via the AUR).


jaspervdj:
Thanks!

PR #688, opened 2019-02-21
Add Nix Package

BenBals:
Unfortunately there is no detail page for the Nix package as I know, so I put a link to the Nix package search and the install command, but I am grateful for suggestions as well.


jaspervdj:
Thanks!

PR #691, opened 2019-03-03
Bumps pandoc upper bound

jkachmar:
This PR bumps the upper bound for `pandoc` to allow `hakyll` to build with `pandoc-2.7`, which was released today.

I've built `hakyll` locally and run tests with the `usePandoc` flag and everything seems to work.

### DIFF
--- a/hakyll.cabal
+++ b/hakyll.cabal
@@ -228,7 +228,7 @@ Library
     Other-Modules:
       Hakyll.Web.Pandoc.Binary
     Build-Depends:
-      pandoc          >= 2.0.5    && < 2.7,
+      pandoc          >= 2.0.5    && < 2.8,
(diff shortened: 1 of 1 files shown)


rpearce:
@jkachmar Looks like there's a dependency issue in circleci around skylighting?
```
Error: While constructing the build
plan, the following exceptions were
encountered:

In the dependencies for pandoc-2.6:
    skylighting-0.7.5 from stack
                      configuration does
                      not
                      match >=0.7.6 && <0.8 
                      (latest matching
                      version is 0.7.7)
needed due to hakyll-4.12.5.1
               -> pandoc-2.6
```
@jaspervdj is there any help we could get to push these changes through?

jaspervdj:
Superseded by #703 -- thanks for the PR though!

PR #695, opened 2019-03-15
Added example

hackeryarn:


jaspervdj:
@hackeryarn Happy to add this example but the source link doesn't seem correct?

hackeryarn:
@jaspervdj Sorry about that mistake. It should be correct now.

jaspervdj:
Thanks!

PR #696, opened 2019-03-16
robertwpearce: hakyll pt 5 tutorial source file

rpearce:
Post: https://robertwpearce.com/hakyll-pt-5-generating-custom-post-filenames-from-a-title-slug.html


rpearce:
@jaspervdj I experienced the issue causing the build to fail locally, as well.

PR #705, opened 2019-05-19
Dependency error on `stack install hakyll`

### LINKED ISSUE #684: Dependency error when running `stack install hakyll`
I'm seeing the following lrucache dependency error when trying to install Hakyll:

```
stack install hakyll
Writing implicit global project config file to: /home/peter/.stack/global-project/stack.yaml
Note: You can change the snapshot via the resolver field there.
Using latest snapshot resolver: lts-13.6
Downloaded lts-13.6 build plan.    
Preparing to install GHC to an isolated location.
This will not interfere with any system-level installation.
Downloaded ghc-8.6.3.                                      
Installed GHC.                                                                 

Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for hakyll-4.12.5.1:
    lrucache must match >=1.1.1 && <1.3, but the stack configuration has no specified version (latest matching version is 1.2.0.1) needed since hakyll is a build target.

Some different approaches to resolving this:

  * Consider trying 'stack solver', which uses the cabal-install solver to attempt to find some working build configuration. This can be convenient when dealing with many
    complicated constraint errors, but results may be unpredictable.

  * Recommended action: try adding the following to your extra-deps in /home/hello/.stack/global-project/stack.yaml:

lrucache-1.2.0.1@sha256:18fc3d7052012c7ab3cd395160f34b53c5e1ec5379cc45185baf35b90ffadc2e

Plan construction failed.
```

I was able to work around this by specifying the lrucache dependency in /home/hello/.stack/global-project/stack.yaml, as the error message suggested, but that shouldn't be necessary.

I also feel like I'm not really grokking the dependency error, since Hakyll has a stack.yml configuration file containing an entry for lrucache which falls within the range defined in hakyll.cabal. Is it possible this file is being ignored by `stack install ...`?

### Environment
Stack: Version 1.9.3, Git revision 40cf7b37526b86d1676da82167ea8758a854953b (6211 commits) x86_64 hpack-0.31.1 (newly installed)
OS: Debian 9.7

AlainODea:
- Set explicit resolver to latest Stackage LTS with hakyll included

Fix #684

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -12,4 +12,4 @@ You can install this library using cabal:
 
 Or using stack:
 
-    stack install hakyll
+    stack install hakyll --resolver lts-12.26
--- a/web/index.markdown
+++ b/web/index.markdown
@@ -32,7 +32,7 @@ and TeX support, including syntax highlighting and other goodies.
 # Getting Started
 
 You can get the latest version from hackage using `cabal install hakyll`, or
-using [stack] by using `stack install hakyll`. Then, you can:
+using [stack] by using `stack install hakyll --resolver lts-12.26`. Then, you can:
 
 [stack]: http://www.haskellstack.org/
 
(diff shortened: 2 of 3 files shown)


AlainODea:
Not really merge-worthy, but useful as a workaround for those curious or impacted.

PR #708, opened 2019-05-29
Allow for multiple categories for getCategory with tests.

dmp1ce:
I didn't know what exactly `getCategory` was supposed to return so I wrote some tests to document the functionality and changed the implementation slightly to fit the tests.

### DIFF
--- a/hakyll.cabal
+++ b/hakyll.cabal
@@ -254,6 +254,7 @@ Test-suite hakyll-tests
     Hakyll.Web.CompressCss.Tests
     Hakyll.Web.Html.RelativizeUrls.Tests
     Hakyll.Web.Html.Tests
+    Hakyll.Web.Tags.Tests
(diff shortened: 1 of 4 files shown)


jaspervdj:
Hi @dmp1ce, thanks for the PR!

This is not really how I think about categories -- the original motivation is that Tags is a many-to-many organization of posts (each post can have multiple tags, and tags appear on multiple posts) whereas Category is a simpler hierarchical model (each post belongs in a Category).

_If_ we were to extend how category works; I think we would need to create a proper tree where categories can have subcategories, but I'm not sure if that's something we want to do.

For now, maybe we can restore the old behavior and just get the tests merged in?  Sorry for the confusion -- any additional documentation around category/tags would also be more than welcome.

dmp1ce:
Let me know if https://github.com/jaspervdj/hakyll/pull/723 works for you.

Issue #714, opened 2019-07-23
Parallel Compilation Support


jhmcstanton:
It would be nice if the compiler could compile `[Rule a]` in parallel for sites that are CPU intensive to build. My specific use case is compressing images - after adding a `compressJpgCompiler` to my site the build time jumped into multiple minutes.

jhmcstanton:
Left some comments on the pull request

PR #715, opened 2019-07-23
(Not ready for merge) Proposal for #714: Parallelizing compiler

jhmcstanton:
Note: no new tests added, but existing tests all pass.

For my use case this worked well, the build went from 4:04 minutes to 2:26 minutes. This was with 99 images marked as separate `Rule`s, with the remainder of my site on a single thread, and running with 4 capabilities. 

If you're OK with this approach then I can clean this up a bit.

### DIFF
--- a/lib/Hakyll/Commands.hs
+++ b/lib/Hakyll/Commands.hs
@@ -1,16 +1,20 @@
  --------------------------------------------------------------------------------
 -- | Implementation of Hakyll commands: build, preview...
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP        #-}
+{-# LANGUAGE Rank2Types #-}
 module Hakyll.Commands
     ( Check(..)
     , build
+    , buildPar
     , check
     , clean
     , preview
     , rebuild
+    , rebuildPar
     , server
     , deploy
     , watch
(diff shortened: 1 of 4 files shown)


jaspervdj:
I think it would be good to (finally) have parallel compilation in Hakyll.  However, I have a slight preference for a traditional "jobs" approach, where we spawn N threads (possibly using the `async` library since it makes things a lot simple) and these N "workers" take items and complete them.  Would you be open to switching to that?

jhmcstanton:
I'm up for any approach really, especially if it simplifies the implementation.

Were you thinking to focus more on [`chase`](https://github.com/jaspervdj/hakyll/blob/master/lib/Hakyll/Core/Runtime.hs#L174) in the Runtime? Or something else?

jhmcstanton:
Just a quick update: been going down the route of async-ifying `chase`. Have it compiling and passing all tests, but I'm currently bumping into some exceptions caused by locked files when running my own site.

jhmcstanton:
@jaspervdj the new pull request using async can be found in #725

LaurentRDC:
This PR was superseeded by #844 and can therefore be closed.

PR #722, opened 2019-08-20
Better error messages

jaspervdj:
This resumes the excellent work done in #462


lib/Hakyll/Core/Compiler.hs
```diff
@@ -177,12 +185,37 @@ cached name compiler = do
...
+noResult :: String -> Compiler a
+noResult = compilerNoResult . return
+
+
+--------------------------------------------------------------------------------
+-- | Prepend an error line to the error, if there is one.  This allows you to
+-- add helpful context to error messages.
+--
+-- @since 4.13.0
+prependErrorMessage :: String -> Compiler a -> Compiler a
```
jaspervdj: I'm not too sure of the name of this function.  I think something like `withErrorContext` would be good but that is confusing since it overloads `Context` which is already used in templates...

jaspervdj:
@bergus I'm pretty keen to merge this in soon.  Do you think you can give it another review?  I made a few changes to your code -- most of it just a number of renames, but I also changed the new error type to use `NonEmpty` in case there are any actual failures.

bergus:
Sure, I'll have a look. Haven't worked with Hakyll for some time now, though a project relying on #620 is still in my backlog...

lib/Hakyll/Core/Compiler.hs
bergus: I see. What about `withErrorMessage`?

lib/Hakyll/Core/Compiler.hs
```diff
@@ -193,20 +194,23 @@ unsafeCompiler = compilerUnsafeIO
...
 -- "Hakyll.Web.Template" macro, and alternative
 -- 'Hakyll.Web.Template.Context.Context's are tried
--- 
--- @since 4.12.0
-failBranch :: String -> Compiler a
-failBranch = compilerFailBranch . return
+--
+-- @since 4.13.0
+noResult :: String -> Compiler a
+noResult = compilerNoResult . return
```
bergus: I like it!

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -112,18 +114,19 @@ instance Monoid CompilerWrite where
...
+data CompilerErrors a
+    -- | One or more exceptions occured during compilation
+    = CompilationFailure (NonEmpty a)
+    -- | Absence of any result, most notably in template contexts.  May still
+    -- have error messages.
+    | CompilationNoResult [a]
     deriving Functor
 
 
 -- | Unwrap a `Reason`
```
bergus: Update docs

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -132,7 +135,7 @@ data CompilerResult a
     = CompilerDone a CompilerWrite
     | CompilerSnapshot Snapshot (Compiler a)
     | CompilerRequire (Identifier, Snapshot) (Compiler a)
-    | CompilerError (Reason [String])
+    | CompilerError (CompilerErrors String)
```
bergus: This is why I did use the name `Reason`, `(Compiler)Errors` looks just weird here.

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -112,18 +114,19 @@ instance Monoid CompilerWrite where
...
+    | CompilationNoResult [a]
     deriving Functor
 
 
 -- | Unwrap a `Reason`
-getReason :: Reason a -> a
-getReason (CompilationFailure x)  = x
-getReason (NoCompilationResult x) = x
+compilerErrorMessages :: CompilerErrors a -> [a]
+compilerErrorMessages (CompilationFailure x)  = NonEmpty.toList x
```
bergus: What's your specific reasoning to avoid empty lists here? And why did `compilerThrow` change to produce a `CompilationNoResult` if no reason value is given?
(Not assuming it's a bad thing, I just don't understand the idea behind it).

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -112,18 +114,19 @@ instance Monoid CompilerWrite where
...
 --------------------------------------------------------------------------------
 -- | Distinguishes reasons in a 'CompilerError'
-data Reason a
-    -- | An exception occured during compilation
-    = CompilationFailure a
-    -- | Absence of any result, most notably in template contexts
-    | NoCompilationResult a
+data CompilerErrors a
+    -- | One or more exceptions occured during compilation
+    = CompilationFailure (NonEmpty a)
```
bergus: What's your specific reasoning to avoid empty lists here? And why did `compilerThrow` change to produce a `CompilationNoResult` if no reason value is given?
(Not assuming it's a bad thing, I just don't understand the idea behind it).

lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -266,21 +272,25 @@ compilerUnsafeIO io = Compiler $ \_ -> do
...
+-- | Throw errors in the 'Compiler'.
+--
+-- If no messages are given, this is considered a 'CompilationNoResult' error.
+-- Otherwise, it is treated as a proper compilation failure.
 compilerThrow :: [String] -> Compiler a
-compilerThrow = compilerResult . CompilerError . CompilationFailure
-
+compilerThrow errs = compilerResult $ CompilerError $ case errs of
+    []       -> CompilationNoResult []
+    (e : es) -> CompilationFailure (e :| es)
```
bergus: I'd prefer `maybe (CompilationNoResult []) CompilationFailure $ NonEmpty.nonEmpty errs`

lib/Hakyll/Web/Template/Internal.hs
```diff
@@ -203,12 +204,16 @@ applyTemplate' tes context x = go tes
 
     ----------------------------------------------------------------------------
 
-    applyStringExpr wrap msg expr = mapError wrap (applyExpr expr) >>= getString
+    applyStringExpr :: Maybe String -> String -> TemplateExpr -> Compiler String
+    applyStringExpr mbPrepend msg expr =
+        prepending (applyExpr expr) >>= getString
```
bergus: I'm not sure any more why I passed `mapError` into `applyStringExpr` - just to prevent code duplication? Or really to not get the error message when `getString` fails, but only when `applyExpr` fails? Could've been a bug, could've been intentional.
So maybe we should just call `prependErrorMessage` at the call site, simplifying `applyStringExpr`.

lib/Hakyll/Core/Compiler/Internal.hs
bergus: Also seems to make the `Alternative.<|>` implementation more complicated

jaspervdj: It's to prevent us from shooting ourselves in the foot -- if there's no error, we shouldn't get into a `CompilationFailure` scenario.  This is not true for the other case where we have `empty`.

lib/Hakyll/Web/Feed.hs
```diff
@@ -77,15 +79,15 @@ data FeedConfiguration = FeedConfiguration
 
 --------------------------------------------------------------------------------
 -- | Abstract function to render any feed.
-renderFeed :: String                  -- ^ Default feed template
-           -> String                  -- ^ Default item template
+renderFeed :: Item String             -- ^ Default feed template
+           -> Item String             -- ^ Default item template
```
jaspervdj: This is my remaining nitpick about the PR -- changing this breaks backwards-compatibility, which is fine, since this is a relatively rarely used function so it's worth it.  However, if we do change it, we should change it to the "right" type which is `Template` or `Item Template` rather than `Item String`.

lib/Hakyll/Web/Template/Internal.hs
bergus: Having another look at the code again, I do remember now: it was intentional. By nesting the `withErrorMessage` calls, one does get told the error location twice in the "stack trace":
```
Hakyll.Web.Template.applyTemplate: expected string but got xy for expr '$…$'
In expr '$…$'
In …
```
Sorry for a) leading you astray with my comment b) not writing test cases for this...

lib/Hakyll/Core/Compiler/Internal.hs
bergus: After familiarising myself with the code again, I would heavily prefer
```hs
-- | Put a 'CompilerError' with  multiple error messages as 'CompilationFailure'
compilerThrow :: NonEmpty String -> Compiler a
compilerThrow = compilerResult . CompilerError . CompilationFailure
```
The `compilerThrow` function is an internal function (only available on `Compiler.Internal`) so we can easily "break" it by changing the type. (IIRC it was earlier used in `Context` or so for transforming errors, but has been replaced there with more sensible alternatives). It never was meant to be called with an empty list, and neither in `fail` nor in `<|>` that could happen.
By having `compilerThrow :: NonEmpty String -> Compiler a`, we don't need to convert back and forth between lists and non-empty lists in `<|>`.

Where it does matter, and where I agree with you that we should prevent anyone from shooting themselves in the foot, is `throwError` in the `MonadError [String]` instance. Let's do
```
throwError errs = maybe (compilerNoResult errs) compilerThrow $ NonEmpty.nonEmpty errs
```

lib/Hakyll/Web/Feed.hs
bergus: `Template` is fine, but it shouldn't be a template constructed by `readTemplate` which doesn't know about the origin of the template file (it just assumes a string literal), which is important for error messages. Doing the `compileTemplateItem` did allow me to use the `Item`s path for the template, which I had explicitly added, without having to use any of the `Template.Internal` stuff to construct this "embedded `Template`".

jaspervdj: @bergus That makes a lot of sense -- what do you think of just providing `embedTemplate` from the template modules then?  It has the filepath that way, and simplifies the Feed code: https://github.com/jaspervdj/hakyll/pull/722/commits/f96b99e963b170ab15fe99341cdc7b2b9543dac4

lib/Hakyll/Web/Template.hs
```diff
@@ -148,8 +149,30 @@ module Hakyll.Web.Template
...
+-- Example:
+--
+-- > myTemplate :: Template
+-- > myTemplate = $(embedTemplate "test.html")
+embedTemplate :: FilePath -> Q Exp
+embedTemplate filePath = [|
+    let source = T.unpack $ T.decodeUtf8 $(embedFile filePath) in
+    case parseTemplateElemsFile filePath source of
+        Left  err -> error err
+        Right tpl -> template filePath tpl |]
```
bergus: Ah, this is how to do it properly with Template Haskell - thanks!

PR #726, opened 2019-08-25
Made inner loops have access to outer loop values

ilyakooo0:
If my outer loop in a template produces a value that I want to use in my inner loop, I get a `Missing field` error. 

This gives inner loops access to outer loop values and allows inner loops to override them.

### DIFF
--- a/lib/Hakyll/Web/Template/Context.hs
+++ b/lib/Hakyll/Web/Template/Context.hs
@@ -272,8 +272,8 @@ titleField = mapContext takeBaseName . pathField
 --
 -- As another alternative, if none of the above matches, and the file has a
 -- path which contains nested directories specifying a date, then that date
--- will be used. In other words, if the path is of the form 
--- @**//yyyy//mm//dd//**//main.extension@ . 
+-- will be used. In other words, if the path is of the form
+-- @**//yyyy//mm//dd//**//main.extension@ .
 -- As above, in case of multiple matches, the rightmost one is used.
 
 dateField :: String     -- ^ Key in which the rendered date should be placed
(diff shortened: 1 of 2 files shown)


jaspervdj:
This looks good, thanks for the PR!  Is it possible to update this with:
1.  An additional comment in `Hakyll.Web.Template.Context.listField`, explaining how the contexts are combined
2.  A test that reads a value from the outer context

ilyakooo0:
Sure

bergus:
I like it, had to do something quite similar in https://github.com/jaspervdj/hakyll/pull/620/commits/f49d972ccae47274d7f8d101237e2a480145ae49 (part of #620). Maybe this `combineContexts` function shouldn't be just local, but exposed in `Context.hs`?
Maybe the `Context` data type needs some general reworking though, if it is supposed to hold some information about an `Item`.

ilyakooo0:
Yeah, I thought about exporting it. I thought it wasn't useful enough.

It sounds like a good idea. Now.

I'm a little overwhelmed right now, will finish the pull request in a couple days.

bergus:
I guess for the API I would prefer a simple
```hs
bindItem :: Context a -> Item a -> forall b. Context b
-- or
bindContext :: Item a -> Context a -> forall b. Context b
```
(not sure which one is better?) over a three-parameter and non-descriptive `combineContexts` function.

You could then write
```hs
ListField c xs -> do
    sep <- maybe (return "") go s
    bs  <- mapM (applyTemplate' b $ c <> bindItem context x) xs
    return $ intercalate sep bs
```

PR #736, opened 2019-10-29
don't require email address for feeds

robx:
(I'm not sure this is the best way to do this, or it's even a good idea to make this optional, feel free to close.)

I found myself wanting to generate an atom.xml without providing an email address. According to [RFC 4287](https://www.ietf.org/rfc/rfc4287.txt), the `<email>` tag is optional.

This changes `feedAuthorEmail` from `String` to `Maybe String`, and updates the template accordingly.


lib/Hakyll/Web/Feed.hs
```diff
@@ -120,6 +120,10 @@ renderFeed feedTpl itemTpl config itemContext items = do
             StringField s -> return s
             _             -> fail "Hakyll.Web.Feed.renderFeed: Internal error"
 
+    emailField = case feedAuthorEmail config of
+        Nothing    -> missingField
+        Just email -> constField "authorEmail" email
```
robx: Would you prefer a new helper

```
optField : String -> Maybe String -> Context a
```

used as `optField "authorEmail" (feedAuthorEmail config)`?

puffnfresh (approved):

jaspervdj:
I'm generally fine with this idea but I've tried really hard to keep breaking changes at bay.  Should we consider just using the empty string as empty?  (I sort of hate the idea, but I also hate the idea of breaking everyone's blogs 😔)

puffnfresh:
@jaspervdj sounds good to me

PR #737, opened 2019-10-29
update stack.yaml to lts-14.12

robx:
When building locally I saw that hakyll builds fine against current lts without the extra-deps. No idea if you want this.


jaspervdj:
Thanks!  I bumped it myself in another commit.

PR #741, opened 2019-11-25
Flush log before starting web server

robx:
This fixes out of order console messaging in the case of errors
binding to the configured address, though the message order is
still a bit weird.

Before:

```
$ stack exec site watch
Initialising...
  Creating store...
site: Network.Socket.bind: resource busy (Address already in use)
  Creating provider...
Listening on http://127.0.0.1:8000
```

After:

```
$ stack exec site watch
Listening on http://127.0.0.1:8000
Initialising...
  Creating store...
  Creating provider...
site: Network.Socket.bind: resource busy (Address already in use)
```


jaspervdj:
Thanks!

PR #753, opened 2020-02-15
Remove constant field title for index.html in the example site

L-TChen:
In the default site generated by Hakyll, `index.html` has specified the filed `title` in its front matter, but it will always been overridden by `constField "title" "Home"` in `site.hs`.

It is a very minor issue, but it may cause confusion to newcomers.


jaspervdj:
Thanks!

PR #762, opened 2020-03-19
Add feed context argument for functions of Hakyll.Web.Feed

gtgteq:
to make more flexible.

More concrete situation is that, to cancel the double slash which link of both items and feed contains, I want to overwrite `url` context.


gtgteq:
I made commit in my master branch. closing.

Issue #764, opened 2020-03-29
Build fails on GHC 8.8.3


emlautarom1:
**TLDR:** Build fails on `ghc 8.8.3 ` with the following error:

```
lib\Hakyll\Commands.hs:102:9: error:
    • No instance for (MonadFail ((->) IOError))
        arising from a use of ‘fail’
    • In a stmt of a 'do' block:
        fail
          $ "Hakyll.Commands.watch: Could not start update watching "
              ++ "thread. Did you compile with -threaded flag?"
      In the second argument of ‘($)’, namely
        ‘do fail
              $ "Hakyll.Commands.watch: Could not start update watching "
                  ++ "thread. Did you compile with -threaded flag?"’
      In a stmt of a 'do' block:
        catchIOError (void $ forkOS $ watchUpdates conf update)
          $ do fail
                 $ "Hakyll.Commands.watch: Could not start update watching "
                     ++ "thread. Did you compile with -threaded flag?"
    |
102 |         fail $ "Hakyll.Commands.watch: Could not start update watching " ++
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
cabal.exe: Failed to build hakyll-4.13.2.0 (which is required by exe:hakyll-init from hakyll-4.13.2.0). See the build log above for details.
```

System specs:
- **ghc**: 8.8.3
- **cabal**: 3.0.0.0
- **OS**: Windows 10 v1909

This might be related to the [MonadFail](https://downloads.haskell.org/ghc/8.8.1/docs/html/users_guide/8.8.1-notes.html#base-library)  changes. It seems like it was discussed and solved on #754 but maybe something has gone under the radar.

emlautarom1:
**Update:** This seems like a Windows issue only. The fix is very small, so I've already made a pull request (#765).

PR #765, opened 2020-03-29
Fix for #764 (Windows issue only)

emlautarom1:
Fixes compile-time error related to `MonadFail` on Windows PCs.
See #764 for more details


jaspervdj:
Thanks!

Issue #766, opened 2020-03-29
hakyll build fails with "cache error" on "templates/default.html"


emlautarom1:
System specs:
- **ghc**: 8.8.3
- **cabal**: 3.0.0.0
- **OS**: Windows 10 v1909

Steps to reproduce:

```
mkdir HakyllSite
hakyll-init -f HakyllSite
cd HakyllSite
cabal run exe:site -- build
```

The output I get is the following:

```
Initialising...
  Creating store...
  Creating provider...
  Running rules...
Checking for out-of-date items
Compiling
  [ERROR] Hakyll.Core.Compiler.Require.load: templates\default.html (snapshot 
_final) was not found in the cache, the cache might be corrupted or the item you are referring to might not exist
```

I can confirm that the `templates\default.html` file exists, so the issue must be somewhere else.

emlautarom1:
**Note:** Deleting the file `templates\default.html` makes no difference to the error, so I think the root of the problem could be on the early `fail notFound`:

https://github.com/jaspervdj/hakyll/blob/b08dd317a381a0436607b44257c487b1930dcc72/lib/Hakyll/Core/Compiler/Require.hs#L55-L63

LaurentRDC:
I'm also encountering this problem right now, on every template file.

LaurentRDC:
I'm getting the same type of error for every template. I am also on Windows.

This error did not occur with Hakyll 4.12.4.0 (stackage lts-12.26). The error occurs when I update to Hakyll 4.13.3.0 (stackage nighlty-05-22-2020).

Looking at the changelog, there is this:

> # Hakyll 4.13.1.0 (2020-02-26)
(...)
> * Fix file path compatibility with Windows (by Hexirp)
(...)

I wonder if that's the source of the error?

LaurentRDC:
Ok I've looked into it a little, and I found that all files that are not in the current working directory are ignored.

For the `hakyll-init` site, this includes files in `posts/`, `templates/`, `css/`, etc.

I'll keep looking.

jaspervdj:
@LaurentRDC thanks for looking into this!  Let me know if there’s anything I can do.

LaurentRDC:
So I have found the source of the problem. Pattern matching (via the functions `Hakyll.Core.Identifier.Pattern.matches` and `Hakyll.Core.Identifier.Pattern.capture` does not work on Windows. This is probably due to the difference between path separators on Windows vs. Posix.

[...]

PR #778, opened 2020-05-26
Fix for filepaths matching on Windows.

LaurentRDC:
This pull request aims to fix issue #766 , where `Rules` were not firing properly on Windows.

Files on Windows were not being captured properly by `Hakyll.Core.Identifier.Pattern.capture`. To fix this, all paths becoming `Identifier`s are being normalized to have the Windows path separator `\\`. This requires modification of `Hakyll.Core.Identifier.Pattern.fromGlob` function. Also, regex matching on Windows paths needed to change a little.

After this fix, the test site produced by `hakyll-init` compiles properly.

I also want to point out that trying to glob `\\*.jpg` is invalid on Windows, so I removed this test case. This also simplifies the parsing of glob patterns.


jaspervdj:
Thanks, this looks good!  Thanks for updating those tests as well!

PR #780, opened 2020-06-04
Allow specifying a metadata provider

### LINKED ISSUE #643: Add Pandoc Meta data to Context
One interesting feature of ReST is that you can define document variables and meta data directly inside the document, without having to rely on some additional YAML header.

For example, the title of the document and its eventual subtitle can be inferred from the first headers of the file (see [Markup Specification (Document Structure)][document]),
and if the first non-comment element is a definition list, its fields update the bibliographic information of the document (see [Markup Specification (Bibliographic Fields)][bibliography]).

[document]: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#document-structure
[bibliography]: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#bibliographic-fields

Although the ReST Parser of Pandoc is far from perfect (it does not support custom directives or roles, and the Pandoc AST is quite restrictive), it does implement the aforementioned features, _in standalone mode_, and populates the Meta information of the Pandoc document.

```rst
Document Title
==============

Subtitle
--------

:author: flupe
```
However, it seems as though the Pandoc compilers provided by Hakyll **completely ignore the meta information** of the parsed Pandoc documents.

I don't really know if other markup languages supported by Pandoc also populate the meta information, but I do think it would be useful to provide an easier way to inject this meta information into Hakyll contexts.

-----------------------

For a custom site I've just set up, it is somehow working a little. Here is the relevant part ([source]):
```haskell
 match "posts/*" $ do
    route $ setExtension "html"

    let ropts = defaultHakyllReaderOptions { readerStandalone = True }
        wopts = defaultHakyllWriterOptions

    compile $ do
        document <- getResourceBody >>= readPandocWith ropts

        let
            Pandoc meta _ = itemBody document

            inlinesToString :: [Inline] -> String
            inlinesToString inlines =
                concatMap inlineToString inlines
                where
                    inlineToString (Str a) = a
                    inlineToString (Space) = " "

            extractMeta :: String -> MetaValue -> Context a
            extractMeta name metavalue =
                case metavalue of
                    MetaInlines inlines -> mkField $ inlinesToString inlines
                    _ -> mempty
                where
                    mkField = field name . const . return

            ctx :: Context String
            ctx = foldMapWithKey extractMeta (unMeta meta)
                <> postCtx

        writePandocWith wopts document
            &   loadAndApplyTemplate "templates/post.html" ctx
            >>= loadAndApplyTemplate "templates/default.html" ctx
            >>= relativizeUrls
```
Essentially, we:
1. Parse the post and get an `Item Pandoc`
2. Create a new context `ctx` populated with the metadata from the document.
3. Render the post with templates using the new context we created.

However, this mechanism in ReST implies that you actually need to parse the document to get the context, and I have no idea how to make this work well with other rules such as creating and `index.html` or `archives.html` page (without having to parse again for each rule).

-------------------------

In conclusion, here's what I am suggesting:
- adding a proper way to transform Pandoc Meta into an Hakyll Context.
- a way to associate this information and posts inside snapshots?

[source]: https://github.com/flupe/site/blob/master/site.hs#L41-L85

ip1981:
In addition to parsing metadata header & metadata files it will allow any custom routing to extract metadata from the resource files, for example with Pandoc. Closes #643.

Feel free to update it to make it more idiomatic and whatnot.

Example:

A simple Pandoc provider for LaTeX
```haskell
module PandocMetadata
  ( pandocMetadata
  ) where

import Hakyll

import qualified Data.Aeson.Types as AT
import qualified Data.HashMap.Strict as HM
import qualified Data.Text as T
import qualified Data.Text.IO as TIO
import System.FilePath (takeExtension)
import qualified Text.Pandoc.Class as PDC
import qualified Text.Pandoc.Definition as PDD
import qualified Text.Pandoc.Options as PDO
import qualified Text.Pandoc.Readers as PDR
import qualified Text.Pandoc.Writers as PDW

pandocMetadata :: FilePath -> IO Metadata
pandocMetadata file = do
  PDD.Pandoc meta _ <- parseFile file
  return $
    HM.fromList .
    map (\(f, t) -> (T.pack f, AT.String t)) .
    filter (not . T.null . snd) .
    map (\(f, ex, wr) -> (f, inlinesTo wr (ex meta))) $
    [ ("title", PDD.docTitle, PDW.writePlain)
    , ("titleHtml", PDD.docTitle, PDW.writeHtml5String)
    , ("date", PDD.docDate, PDW.writePlain)
    ]

parseFile :: FilePath -> IO PDD.Pandoc
parseFile file = do
  cnt <- TIO.readFile file
  case PDC.runPure $ PDR.readLaTeX PDO.def cnt of
    Right t -> return t
    Left e -> error $ "Pandoc read failed: " ++ show e

inlinesTo ::
     (PDO.WriterOptions -> PDD.Pandoc -> PDC.PandocPure T.Text)
  -> [PDD.Inline]
  -> T.Text
inlinesTo wr ill =
  case PDC.runPure . wr PDO.def $ doc of
    Right t -> t
    Left e -> error $ "Pandoc write failed: " ++ show e
  where
    doc = PDD.Pandoc PDD.nullMeta [PDD.Plain ill]
```

Usage:
```haskell
main :: IO ()
main = do
  opts <-
    O.execParser
      (O.info
         (parseOptions <**> O.helper)
         (O.fullDesc <> O.header "Static site compiler"))
  let conf =
        Conf.defaultConfiguration
          { Conf.destinationDirectory = outDir opts
          , Conf.providerDirectory = srcDir opts
          , Conf.storeDirectory = cacheDir opts
          , Conf.tmpDirectory = cacheDir opts </> "tmp"
          , Conf.provideMetadata = pandocMetadata
          }
  log <-
    Logger.new
      (if verbose opts
         then Logger.Debug
         else Logger.Message)
  case command opts of
    Build -> Cmd.build conf log rules >>= exitWith
    Clean -> Cmd.clean conf log
    Check chk -> Cmd.check conf log chk >>= exitWith
```


jwiegley:
This is exactly what I needed today, in order get metadata from Org-mode properties, rather than a Yaml block.

Alf0nso:
Any updates on this? I would love for the possibility to just use org-mode metadata, or for a way to specify the format of my metadata...

jwiegley:
I've been using this for a while now, and would love to see it added into the main release.

lib/Hakyll/Core/Provider.hs
```diff
@@ -24,20 +25,29 @@ module Hakyll.Core.Provider
...
+-- | Create a resource provider with the void metadata provider
 newProvider :: Store                   -- ^ Store to use
             -> (FilePath -> IO Bool)   -- ^ Should we ignore this file?
             -> FilePath                -- ^ Search directory
             -> IO Internal.Provider    -- ^ Resulting provider
-newProvider store ignore directory = do
+newProvider store ignore = newProvider' store ignore (const $ return mempty)
+
+-- | Create a resource provider
+newProvider' :: Store                   -- ^ Store to use
```
jaspervdj, line 43: Even though it's a bunch of boilerplate, we should probably add a new `ProviderOptions` together with `defaultProviderOptions` so we can add new fields in a backwards-compatible way.

PR #782, opened 2020-06-11
Fix GitHub workflow scripts

L-TChen:
It doesn't seem possible to use `strategy.matrix` to specify what OS to run. So, instead, there are two separate jobs for Ubuntu and for Windows. Moreover, the GitHub Action `mstksg/setup-stack` does not support Windows despite what it claims (see [here](https://github.com/mstksg/setup-stack/issues/5)), so it doesn't make sense to use the same steps on different OS environments.

For some compatibility issue, I put an additional `stack-8.8.3.yaml` so that in the future it is easier to run workflows for different GHC versions if you'd like.

Due to a GHC bug, it fails to install Hakyll on Windows for some mysterious reason while building SHA: 

> SHA                              > Access violation in generated code when writing 0x0

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,20 +1,58 @@
-name: 'CI'
-on: ['push']
+name: CI
+on: [push]
+
 jobs:
-  build:
+  Ubuntu:
+    runs-on: ubuntu-latest
+    
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
-      
-    name: ${{ matrix.os }}
-    runs-on: $${{ matrix.os }}
-    
+        ghc-ver: [8.8.3]
+
+    env:
+      ARGS: "--stack-yaml stack-${{ matrix.ghc-ver }}.yaml --no-terminal --system-ghc --fast"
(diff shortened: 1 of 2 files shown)


LaurentRDC:
You can definitely combine all steps for both operating systems (see [here](https://github.com/LaurentRDC/pandoc-plot/blob/599931824f1716de9d0e1738e9b8ff02e15f0727/.github/workflows/ci.yml)). In this case, don't use the `setup-stack` action, but rather the `setup-haskell` action like so:

```yaml
    # https://github.com/actions/setup-haskell
    - uses: actions/setup-haskell@v1.1
      name: Setup Haskell Stack
      with:
        stack-version: "latest"
        enable-stack: true
        stack-no-global: true
```

L-TChen:
Interesting. so `mstksg/setup-stack` can simply be replaced by `actions/setup-haskell` completely. Then it makes sense. 

Do you have an updated script? If so, I’d like to close mine.

LaurentRDC:
Ok I'll update the CI script with what I have described above, and will create a new PR.

PR #779, opened 2020-05-27
Miscellaneous Windows-specific fixes and CI

LaurentRDC:
Hello,

Following #778 , I noticed that a few tests did not pass on Windows. This pull request fixes a few things, including tweaked expected test values (e.g. with newlines `\r\n` instead of `\n`)

I also noticed that `gsubRoute` did not fire properly due to Regex not matching Windows path separators. This also has been fixed.

I noticed that the `toUrl` function also did not always work with Windows file paths, and this has been fixed as well. I added a test to ensure no regressions.

At this time, all tests pass on Windows!

@jaspervdj, what do you think of adding Windows CI to Github Actions? I took care of this in this PR as well. I can commit to fix things specific to Windows for the foreseeable future.

PR #783, opened 2020-06-12
Continuous integration on Windows and Linux

LaurentRDC:
Hello,

This pull request is an extension of #779, following the discussion in #782.

The only change is related to the GH Workflow file, which configures CI on both Linux and Windows. Contrary to #779 , I can confirm that the CI works correctly.

This PR also activates CI builds for other PRs, so that the problem does not repeat :).

One thing to note. As mentioned in #782, CI builds on Windows using GHC 8.8.1 - 8.8.3 do not work reliably. See [here](https://gitlab.haskell.org/ghc/ghc/issues/17599) and [here](https://gitlab.haskell.org/ghc/ghc/issues/17926). Therefore, until fixes are backported from GHC 8.10.*, Windows builds are using GHC 8.6.5. 

BTW, I'm sorry about the large number of commits. I had to validate many small changes by running CI over and over. I tried squashing all the commits, but evidently this is not reflected on GitHub. Don't forget to squash when accepting the PR.


LaurentRDC:
FYI the test that is failing on Windows is related to issues regarding `removeDirectoryRecursive`. See [here (github.com/haskell/win32/pull/129)](https://github.com/haskell/win32/pull/129) for a summary. Basically, trying to remove directories recursively *too fast* may fail with `The directory is not empty`.

Potential solutions are described [here (github.com/haskell/directory/issues/96)](https://github.com/haskell/directory/issues/96), but were never merged.

I am working on a workaround.

LaurentRDC:
Uhh so adding a delay of 100 microseconds when deleting directories quickly on Windows fixes it.

If other people are stumbling on this thread, the fix looks like this:

```haskell
import Control.Exception
import System.Directory   hiding (removeDirectoryRecursive)


removeDirectoryRecursive :: FilePath -> IO ()
removeDirectoryRecursive = retryWithDelay 10 . removePathForcibly


retryWithDelay :: Int -> IO a -> IO a
retryWithDelay i x 
    | i <= 0    = error "retry count must be 1 or more"
    | i == 1    = x
    | otherwise = catch x $ \(_::SomeException) -> threadDelay 100 >> retryWithDelay (i-1) x
```

jaspervdj:
Thanks, this is awesome!

PR #798, opened 2020-10-07
Update a reference to Debian stable

mdimjasevic:
Hakyll has been available in the Debian stable release for a while now so this patch makes it clear one does not need to use Debian unstable to install a package for Hakyll.


Minoru:
Thank you!

PR #855, opened 2021-06-05
Allow demotion of headers by a given amount

ThisFieldWasGreen:
Hello! I found myself in a spot where I needed to _really_ demote some headers, and I thought that other people might find this feature useful.

I've added a new function `demoteHeadersBy` which takes an argument `n > 0` and applies the demotion down no further than `h6`. The existing `demoteHeaders` function is now just an alias of `demoteHeadersBy 1` and retains its existing functionality.

Let me know how this looks and if I need to add or change anything. I didn't find a formal guide for contributing code so I'm just tossing this out there :)

### DIFF
--- a/lib/Hakyll/Web/Html.hs
+++ b/lib/Hakyll/Web/Html.hs
@@ -7,6 +7,7 @@ module Hakyll.Web.Html
 
       -- * Headers
     , demoteHeaders
+    , demoteHeadersBy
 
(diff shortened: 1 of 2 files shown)


Minoru (approved):
Looks great! Thank you very much for the work, and congratulations on your first PR to Hakyll! :tada:

PR #862, opened 2021-07-15
Supporting different field names for tags.

jhmcstanton:
Recently had a situation where we wanted two separate fields for tags, one for a simple tag cloud and another for a more robust index of keywords, and implemented these in our site generator. This just refactors `getTags` and `buildTags` to easily allow users to pick an arbitrary field for tag-like data without needing to peek into the source.

### DIFF
--- a/lib/Hakyll/Web/Tags.hs
+++ b/lib/Hakyll/Web/Tags.hs
@@ -112,11 +112,11 @@ getTags = getTagsByField "tags"
 -- | Obtain tags from a page by name of the metadata field. These can be a list
 -- or a comma-separated string
 getTagsByField :: MonadMetadata m => String -> Identifier -> m [String]
-getTagsByField field identifier = do
+getTagsByField fieldName identifier = do
     metadata <- getMetadata identifier
     return $ fromMaybe [] $
-        (lookupStringList field metadata) `mplus`
-        (map trim . splitAll "," <$> lookupString field metadata)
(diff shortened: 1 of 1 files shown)


lib/Hakyll/Web/Tags.hs
```diff
@@ -143,6 +150,11 @@ buildTags :: MonadMetadata m => Pattern -> (String -> Identifier) -> m Tags
 buildTags = buildTagsWith getTags
 
 
+--------------------------------------------------------------------------------
```
Minoru, line 153: Sorry, this is such a nitpick, but I think this function should be dropped. If we keep it, a question arises why there are no `…ByField` versions of other functions that call `getTags`, like `tagsField`. I fear that if we go down that path, we'll end up with an exponential growth.

It's easy enough to implement `buildTagsByField` and `tagsFieldByField` on the user side, so let them do it I think.

Minoru (changes requested):

lib/Hakyll/Web/Tags.hs
jhmcstanton: No worries, that' a fair point. I had a similar concern but had already implemented this one. 

It's removed.

Minoru (approved):
You forgot to remove the function from the export list :P I fixed it and will merge once the CI is through.

Thanks for the work! :+1:

jhmcstanton:
oof, thanks for taking care of that!

PR #883, opened 2021-09-18
Add support for  jupyter notebooks

fedeinthemix:
Pandoc can now handle Jupyter notebooks. Add support for this format to Hakyll.


fedeinthemix:
Actually the only relevant commit is the second one (57a6fda). The first one is my other PR #882 which unfortunately I merged in my master copy before thinking about this one.

Maybe you could cherry-pick it.

Minoru (approved):
Thanks, looks good! I'll squash all your commits so the final diff looks clean. (Just waiting for the CI now.)

PR #921, opened 2022-02-23
web: Fix link to part one of Robert Pearce's tutorial

slotThe:
The URL must have changed at some point, as I can even find some dead links on that person's website.  I rebuid the site with the `buildWebsite` flag and then checked things via `stack exec hakyll-website watch` and it looked fine to me—hopefully that was correct


jaspervdj:
Thanks!

Issue #614, opened 2018-03-04
hakyll can't handle unicode?


davidspies:
Trying to build the default (hakyll-init) site I get this:

```
$ stack exec site build
Initialising...
  Creating store...
  Creating provider...
  Running rules...
Checking for out-of-date items
Compiling
  [ERROR] ./about.rst: hGetContents: invalid argument (invalid byte sequence)

```
It looks like it's a problem with the unicode character (it builds when I remove it):
The line:
`1. Amamus Unicode 碁`

Hakyll and site built with `nix` and `stack --resolver=lts-10.7`

cbzehner:
I don't think Unicode is the root cause here. I have built [Unicode pages](http://www.cbzehner.com/posts/1970-01-01-security-tips.html) successfully with Hakyll.

Could you [check the encoding](https://unix.stackexchange.com/questions/11602/how-can-i-test-the-encoding-of-a-text-file-is-it-valid-and-what-is-it) of your `about.rst` file? It might be in some other encoding like [Big5](https://en.wikipedia.org/wiki/Big5)?

jaspervdj:
This is a `readFile` exception that most commonly pops up if your system locale is set to something like ISO 8859-1 rather than UTF-8.  On Linux this usually means that `LANG` / `LC_ALL` is set incorrectly.

davidspies:
```
$ file -i about.rst 
about.rst: text/plain; charset=utf-8

```

davidspies:
```
$ locale
LANG=en_US.UTF-8
LANGUAGE=en_US
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=

```

However I built with `nix` so maybe it wipes that out?

davidspies:
Oh, yeah:
$ stack --nix exec -- locale
```
LANG=
LC_CTYPE="POSIX"
LC_NUMERIC="POSIX"
[...]

PR #1017, opened 2024-02-06
Add some docs on how to ensure proper locale access in hakyll

why-not-try-calmer:
Consolidating information from:
- https://github.com/NixOS/nix/issues/318
- https://github.com/jaspervdj/hakyll/issues/614

### DIFF
--- a/web/tutorials/01-installation.markdown
+++ b/web/tutorials/01-installation.markdown
@@ -67,12 +67,6 @@ Create the `my-site` directory with the project files inside:
 Now, change into `my-site` directory and run `stack init` to create the
 `stack.yaml` file.
 
-On NixOS you will probably have to add the following lines to this file:
-
-    nix:
-      enable: true
-      packages: [zlib.dev, zlib.out]
-
 The file `site.hs` holds the configuration of your site, as an executable
 haskell program. We can compile and run it like this:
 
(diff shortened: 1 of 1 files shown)


why-not-try-calmer:
Any news about this?

jaspervdj (approved):
I think it's a good addition since this has come up a few times, but it may also be worth adding this to a non-nix-specific place...

why-not-try-calmer:
> I think it's a good addition since this has come up a few times, but it may also be worth adding this to a non-nix-specific place...

Got it, since you have approved the PR would you prefer me to add / update the docs with the more generic case in a separate PR?

jaspervdj:
No real preference, adding it to this PR is slightly easier for me but if you want to get the changes our earlier I can also deploy this as-is and you can do a follow-up (if you want to).

why-not-try-calmer:
> No real preference, adding it to this PR is slightly easier for me but if you want to get the changes our earlier I can also deploy this as-is and you can do a follow-up (if you want to).

I was about to add it to the FAQ when I saw [this entry](https://jaspervdj.be/hakyll/tutorials/faq.html#hgetcontents-invalid-argument-or-commitbuffer-invalid-argument). Could it suffice?

PR #1044, opened 2024-09-19
Add `boolFieldM` for `Compiler`-dependent bool fields

0xd34df00d:
This is useful if the value of the field depends on item's metadata, for instance, which is best queried in a `MonadMetadata`, which `field'`s type fixes to be a `Compiler`.


LaurentRDC:
Awesome, thank you!

PR #1046, opened 2024-10-05
Update Stack snapshot version in example command.

chungyc:
It seems reasonable to update the Stack snapshot in the example command so that it is no longer 3 major versions behind.


LaurentRDC:
Failure in tests is due to https://github.com/haskell-fswatch/hfsnotify/issues/116, which is unrelated.

PR #1053, opened 2024-11-13
Change location of "Fork me on GitHub" image.

chungyc:
The one on AWS no longer works, and the broken image has been bugging me for a while.  This changes it to the one specified on https://github.blog/news-insights/github-ribbons/.

I would really like to change it to something based on CSS and not an image (in particular, it would be easier to adjust with different color schemes if that is ever supported), but this is the minimum change to fix the broken image.

### DIFF
--- a/web/templates/default.html
+++ b/web/templates/default.html
@@ -16,7 +16,7 @@
     </head>
     <body>
         <!-- Fork me on Github -->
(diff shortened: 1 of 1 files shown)


LaurentRDC:
Thank you!
I don't have the ability to deploy the website though, so I have no idea when this PR will be reflected

PR #1056, opened 2024-11-24
Add option to include WAI middleware for the preview server. 

chungyc:
*   Adds the option to add middleware between the HTTP server and the static file serving to the configuration.

*   Update a number of functions to use the configuration as much as possible, instead of using the configuration to compute some values and then using these as arguments.  The list of parameters was getting pretty long for some of these.

*   Add some more documentation for `Hakyll.Core.Configuration` and organize into sections to make it clearer what may be optional and how to use things.

*   Add a `middlewareRefresh` function for easily adding `Refresh` HTTP headers.  (This is the main motivation I have for this change.)

    Other potential utility middleware that could be added in the future are overriding MIME types (although this one can already be done by using custom server settings) or directing the request to different files depending on HTTP headers such as `Accept-Language`.

After switching windows to manually reload preview pages so many times, I wanted pages to be reloaded automatically.  I was a conflicted between providing a specific configuration field only for enabling automatic refresh or generic field for inserting arbitrary WAI middleware.  I settled on the latter; while it may give _too_ much flexibility and make it possible to use the preview server in ways it really should not be, it can subsume a lot of extra fields that anyone may want to add to the configuration in the future.  And maybe we shouldn't try too hard to prevent people from doing what they want.


LaurentRDC:
I'm not against this in principle, but this PR is quite the breaking change, touching the core `Configuration` type. 

For your particular use case, there are other ways you might try first (e.g. `watch -n 10 'cabal run -- site build'` to build your site every 10 seconds).

We could consider merging this patch once we have a backlog of other breaking changes, to minimize churn

Issue #1052, opened 2024-11-10
Update links to distribution packages


UnlimitedCookies:
Fedoras new link: https://packages.fedoraproject.org/pkgs/ghc-hakyll/ghc-hakyll/
Archlinux's package could also be added: https://archlinux.org/packages/extra/x86_64/haskell-hakyll/

And perhaps documentation could be added how to use Hakyll when it was installed as a system library. One way is to add the runhaskell shebang to the head of the `site.hs` file for example.

dhruvdabhi101:
Hey, I want to work on this if this is open to take.

LaurentRDC:
Of course, go ahead

PR #1057, opened 2024-11-29
Updated links of distribution packages in docs

dhruvdabhi101:
- Changed `installation.makdown` file in `web` directory and changed links in there.


dhruvdabhi101:
> > And according to #1052, I have to add docs on how to run hakyll as system ...
> 
> Do as much as you are comfortable with. We can merge this pull request as-is, and keep issue 1052 open.
> 
> If you DO know how to use a system-installed Hakyll, then this PR would be the right place for it. Up to you

Yeah, let's just merge this, i will get more into that today, and will open another PR for that. Is that fine ?

LaurentRDC:
Perfect, thank you

PR #1058, opened 2025-01-10
`processPandocBiblio`: Honour `nocite` directives

slotThe:
Adds support for pandocs [nocite](https://pandoc.org/chunkedhtml-demo/9.6-including-uncited-items-in-the-bibliography.html) functionality by having an appropriate field in the metadata of the site:

    nocite: |
      @citation1
      @citation2
      …

This adds `pandoc-types` as an explicit dependency, but it was already transient before that, so it's not really new. I also wrote some documentation (in a separate commit, but feel free to squash) so more people can understand how to actually use these functions.

### DIFF
--- a/hakyll.cabal
+++ b/hakyll.cabal
@@ -252,7 +252,8 @@ Library
     Other-Modules:
       Hakyll.Web.Pandoc.Binary
     Build-Depends:
-      pandoc >= 2.11 && < 2.20 || >= 3.0 && < 3.7
+      pandoc >= 2.11 && < 2.20 || >= 3.0 && < 3.7,
+      pandoc-types >= 1.22 && < 1.24
     Cpp-options:
       -DUSE_PANDOC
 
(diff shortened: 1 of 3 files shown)


jaspervdj (approved):
Looks good to me, thanks for adding the docs as well!  Feel free to add a `contribution by` note in the CHANGELOG using your preferred name if you want.

hakyll.cabal
```diff
@@ -252,7 +252,8 @@ Library
     Other-Modules:
       Hakyll.Web.Pandoc.Binary
     Build-Depends:
-      pandoc >= 2.11 && < 2.20 || >= 3.0 && < 3.7
+      pandoc >= 2.11 && < 2.20 || >= 3.0 && < 3.7,
```
LaurentRDC, line 255: [pandoc 2.11](https://hackage.haskell.org/package/pandoc-2.11) through to 2.19 had support for pandoc-types 1.22; is there a reason to be more restrictive on the pandoc-types lower bound?

slotThe: Not really, I just put in whatever `cabal gen-bounds` gave me (which I realise is not a very scientific way to go about things)

LaurentRDC: Understood, please allow pandoc-types-1.22 in this case, such that we don't necessarily (implicitly) restrict the version of pandoc to be < 3

LaurentRDC: Thank you!

LaurentRDC:
Thanks @slotThe ! I will prepare a new release

LaurentRDC:
This feature has been released in [Hakyll 4.16.5.0](https://hackage.haskell.org/package/hakyll-4.16.5.0)

Issue #1062, opened 2025-02-15
Crash with "getDirectoryContents:openDirStream: permission denied (Permission denied)"


wrengr:
The `Hakyll.Core.Util.File.getRecursiveContents` function does not check for permissions before it calls `System.Directory.getDirectoryContents`. Therefore, when encountering directories that it does not have permissions for, it will crash the program with the aforementioned error message.

In particular, this happens when `Hakyll.Core.Runtime.run` calls `Hakyll.Core.Provider.Internal.newProvider` which in turn calls `getRecursiveContents`. Which is to say, every time we start up our hakyll program it will try to read every single thing underneath the current directory — including things like cabal's `dist-newstyle` directory, darcs's `_darcs` directory, and hakyll's own `_cache` and `_site` directories!  This is horribly inefficient, not to mention a security/correctness problem. Those specific directories have strange names, so it's easy enough to filter them out by setting the `shouldIgnoreFile` field of the `Configuration`.  However, because `shouldIgnoreFile` can only look at the basename of each file, it cannot filter out all sorts of other directories the user might want to keep hakyll out of (e.g., I like to have top-level directories with names like 'TODO' or 'BUGS', so I can keep track of notes-to-self; but then maybe I also want my website to have directories with those names). Therefore, it would also be nice to have some mechanism for filtering out directories based on the complete path (relative to `top`); or even better would be to make it a whitelist instead of a blacklist/filter.

-----

I discovered this crash when trying to work around the crash reported in https://github.com/jaspervdj/hakyll/issues/1061. I was hoping to triage that crash by simply removing permissions to the directory where the broken symlinks live, which uncovered the current crash.

These two bugs might be more related that it appears on the surface. That is, if `System.Directory.doesDirectoryExist` returning false is ever interpreted to mean the thing is a (standard) file, then that's a problem: because it returns false for broken symlinks.  I'm not sure if that mistake is actually made anywhere in the codebase, but the two crashes make me wonder how posixly-correct the rest of the file manipulation code is.

LaurentRDC:
There are a few things in this issue.

First, the cause of the crash:[ it is true that permissions are not checked](https://github.com/jaspervdj/hakyll/blob/1112cbe05d869d75dbedd8aeaf3ceda5d9cdbc3d/lib/Hakyll/Core/Util/File.hs#L36); all files in a directory are crawled and returned.

[...]

Issue #1061, opened 2025-02-15
Crash with "getModificationTime:getFileStatus: does not exist (No such file or directory)"


wrengr:
The root cause of this error is whenever there's a broken symlink below the root directory where you're running the hakyll program.

I haven't tracked down what exactly in the code is causing it. But it happens very early on, around the time when `Hakyll.Core.Runtime.run` says it's creating the new provider and store.

PR #1064, opened 2025-02-18
Do not return broken symlinks from getRecursiveContents (#1061)

wrengr:
Also see #1062


wrengr:
Whoops, meant to make a branch for this; will re-file the PR

wrengr:
See https://github.com/jaspervdj/hakyll/pull/1065 instead

PR #1065, opened 2025-02-18
Do not return broken symlinks from getRecursiveContents (#1061, #1062)

wrengr:
This is mostly for fixing #1061 but it also addresses some concerns discussed in #1062


LaurentRDC:
Looks great, thanks! I'll make a new release

LaurentRDC:
Change released in [4.16.6.0](https://hackage.haskell.org/package/hakyll-4.16.6.0)!

PR #1067, opened 2025-03-14
Add support for typst

loutr:
Pandoc supports Typst as an input filetype. This tiny pull request adds the ability for Hakyll to read typst files.

### DIFF
--- a/lib/Hakyll/Web/Pandoc.hs
+++ b/lib/Hakyll/Web/Pandoc.hs
@@ -67,6 +67,7 @@ readPandocWith ropt item =
         MediaWiki          -> readMediaWiki ro
         OrgMode            -> readOrg ro
         Rst                -> readRST ro
+        Typst              -> readTypst ro
         Textile            -> readTextile ro
         _                  -> error $
             "Hakyll.Web.readPandocWith: I don't know how to read a file of " ++
--- a/lib/Hakyll/Web/Pandoc/FileType.hs
+++ b/lib/Hakyll/Web/Pandoc/FileType.hs
@@ -33,6 +33,7 @@ data FileType
     | PlainText
     | Rst
     | Textile
+    | Typst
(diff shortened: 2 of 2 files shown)


LaurentRDC:
Great, thank you!
I'll let the CI check everything before merging. Once that's done, I can make a quick release (probably over the weekend).

LaurentRDC:
Oof, looks like older versions of `pandoc` doesn't support Typst. 

Backwards-compatibility is very important to this project, so we need to support these older GHC versions as much as possible. 

Are you comfortable wrapping your code in CPP such that your change not prevent compilation with older GHCs?

LaurentRDC:
No worries, I can take care of it after merging. I will explain the commit here afterwards

LaurentRDC:

LaurentRDC:
You can see the logic in the commit linked just above this message.

Hakyll supports pandoc 2.11+, but Typst support landed in Pandoc 3.1.3. 

We have two options, then:
* Increase the lower bound on Pandoc, such that Typst is always supported;
* Not support Typst

Neither of these options are great.

There is a third option: use a Haskell language extension called `CPP` (C Pre-Processor), which allows us to conditionally compile part of the code. For example, in the linked commit, you'll see something like:

```haskell
{-# LANGUAGE CPP#-}

...

#if MIN_VERSION_pandoc(3,1,3)
    -- Something where Typst is supported
#else
    -- Something where Typst is not supported
#endif
```

In this way, the compilation of Hakyll will result in different packages whether Pandoc is recent enough or not. This is a trick used throughout the codebase to ensure that Hakyll remains available for people with older versions of GHC.

loutr:
Alright that makes complete sense, I guess I was a bit confused and tired when I answered. Anyway I am still new to the Haskell ecosystem, thank you for taking the time to provide the explanation :)

PR #1069, opened 2025-03-31
Simplify replaceAll and splitAll in Hakyll.Core.Util.String

thomie:
`(=~~)` is polymorphic in its return type. Get first match + before/after directly, instead of splitting manually on match indices. Moreover, since `=~~` uses `'fail'` on lack of match, and `Maybe` is an instance of `MonadFail`, `listToMaybe` isn't necessary.


LaurentRDC:

While we're touching these functions, would you mind adding a few test cases [here](https://github.com/jaspervdj/hakyll/blob/517b5c413f1e7655d471c1c3ae8a365a26db647f/tests/Hakyll/Core/Util/String/Tests.hs#L24)?

thomie:
Here you go. Even found a bug!

PR #1076, opened 2025-08-05
Re-link pandoc in the site to `https://pandoc.org/` 

foxmean:
The previous link is link to the broken site of John McFarlane like the picture below.
image
So, I link to official site instead.


LaurentRDC:
Thanks!

Issue #1077, opened 2025-08-21
Validate generated feed xml?


stuebinm:
I accidentally broke my rss feed recently by having a `&` in a title field, leading to non-well-formed XML in the field (I didn't notice, and it took months for a reader to point this out to me).

This has been discussed before in https://github.com/jaspervdj/hakyll/issues/706 (also https://github.com/jaspervdj/hakyll/pull/750) which suggests always html-escaping the title (and some other) fields, but this was not merged so as to not disallow other html tags. That's fair enough, but also means the original problem is still around.

However: could we instead make `renderRss`/`renderAtom` validate the xml it's produced, and generate an error in case it's not well-formed? This would keep the current flexible behaviour, but in cases like a non-`&amp;`'d "&" would also let authors know their feed's broken.

LaurentRDC:
That's a great suggestion, thanks.
I'm always happy to review pull requests for this. [The `xml` package has a simple `parseXML` function](https://hackage.haskell.org/package/xml-1.3.14/docs/Text-XML-Light-Input.html) that could be used for this purpose

stuebinm:
I took a stab at this, but it seems that `parseXML` is much too lenient to complain about such issues, e.g.:

~~~ghci
ghci> parseXML "a & b"
[Text (CData {cdVerbatim = CDataText, cdData = "a & b", cdLine = Just 1})]
~~~

(there's also a `parseXMLDoc`, but all that it requires is that there's at least a root tag; it needn't even be closed)

However, there's also `renderText`/`renderLBS` from `xml-conduit`, which is likewise already in Hakyll's dependency closure, and which is a little more strict about well-formed XML:

~~~ghci
ghci> parseText def "a & b"
Left 1:1 (0)-1:3 (2): Found content after root element: EventContent (ContentText "a ")
~~~

I've implemented a check using it in https://github.com/jaspervdj/hakyll/pull/1078.

LaurentRDC:
Closed via #1078, released in [Hakyll 4.16.7.0](https://hackage.haskell.org/package/hakyll-4.16.7.0)

PR #1078, opened 2025-08-24
Validate XML of generated Atom/RSS feeds

stuebinm:
Contents of the metadata fields of posts are included verbatim in the generated Atom/RSS xml. This is intentional (we want to allow setting markup tags in e.g. a post's title), but previously made it easy to accidentally have hakyll produce invalid XML in feeds if any field contained invalid XML.

Now instead these will be thrown as a compiler error (using xml-conduit's `parseText` to validate the XML). `xml-conduit` is already in Hakyll's dependency closure, so this adds no new dependency, just makes a formerly implicit one explicit.

One note: if parsing fails, `xml-conduit`'s parse functions will return a `SomeError`. With both `show` and `displayException` these produce strings like `ParseError {errorContexts = ["text content"], errorMessage = "Failed reading: parseTextContent", errorPosition = 12:22 (521)}`, which seem a little ugly, but since they're in a `SomeException`, I'm not sure how to print them any nicer.

I have tested this on my own blog, where it successfully caught https://github.com/jaspervdj/hakyll/issues/1077#issuecomment-3215252152.


LaurentRDC:
Thanks @stuebinm!

I've got two high-level comments:
1. Can you add a test case that catches the problem you experienced?
2. Can you add a method to *bypass* this? This should be turned on by default, but still there should be a way to opt-out of this behavior while it is new

stuebinm:
> As long as there's some way to turn if off, such that no one is blocked for whatever reason

I've also added a note to the top of the module's documentation on this (tho I think the function names should be pretty self-explanatory)

LaurentRDC (approved):
This is great, thank you!

LaurentRDC:
[Released in Hakyll 4.16.7.0](https://hackage.haskell.org/package/hakyll-4.16.7.0)

PR #1084, opened 2025-11-02
Add support for declaring metadata-only dependencies

### LINKED ISSUE #383: How to cache parts of an item?
Hi,
I've got some problems where Hakyll seems to recompile some pages even though their contents did not change. It is right that files they depend on did change, but not the parts (metadata) that are actually used.
Does Hakyll support this use case in its dependency tracking? Is there anything I can do to change my code (e.g. use `cached` somehow)?

Here's a very simplified example of how I create my menus:

``` hs
main = hakyll $ do
  match "pages/**.html" $ do
    route   idRoute
    compile $ do
      page <- getResourceBody
      saveSnapshot "raw" $ itemSetBody () page -- I thought deleting the body would help
      loadAndApplyTemplate "templates/default.html" pageContext page
  match "templates/*" $ compile templateCompiler

allPages :: Compiler [Item ()]
allPages = loadAllSnapshots ("pages/**.html") "raw"

pageContext = field "nav" makeMenu <> defaultContext

makeMenu :: Item String -> Compiler String
makeMenu curPage = do
  pages <- allPages
  menus <- mapM (loadAndApplyTemplate "templates/menuitem.html" menuItemContext) pages
  return $ intercalate "\n" menus
  where
    menuItemContext :: Context ()
    menuItemContext = boolField "active" isActive <> urlField "url" <> metadataField
    isActive menuItem = itemIdentifier curPage == itemIdentifier menuItem
```

My `default.html` template uses `$nav$` somewhere in a sidebar, and my `menuitem.html` template looks somewhat (without linebreaks, actually) like this:

``` html
$if(active)$
  <li class="active"><strong>$title$</strong></li>
$else$
  <li><a href="$url$">$title$</a></li>
$endif$
```

In reality, `pages` undergoes a lot more mangling, and the menu items are actually selected using `functionField`s in a template so that I can order them, and `allPages` does only load pages from a certain directory depending on which menu the page uses, but I think you get the gist. Disclaimer: Above code is untested.

Now when I change the `title` metadata of a page and call `build`, all pages are re-rendered and the menu is updated everywhere. However, if I change only the body of a page, it still **recompiles all pages not only the changed one**. I don't need to re-generate all menus. **How can I accomplish that?**

I guess `Hakyll.Web.Tags` has a similar problem, but I couldn't figure out whether or how it is solved there. Does collecting metadata in the `Rules` monad (and manually creating dependencies) make a difference here?

nmeum:
Presently, it is not possible to only depend on the metadata of an item. This is problematic because it can cause a lot of unnecessary rebuilds when a compiler only needs to run on metadata changes. For example, consider a navigation sidebar that includes all tags of available posts. If this sidebar is included on every page, then we would always rebuild everything when changing the content of any post (even though the metadata tags themselves didn't change!).

To overcome this limitation, we need to distinguish between content and metadata dependencies. This is basically what #383 is about, where @jaspervdj also expressed interest in this feature. However, since nobody has worked on this [since 2015](https://github.com/jaspervdj/hakyll/issues/383#issuecomment-150836917), I decided to give it a shot.

The implementation proposed here is split into two commits, which can be reused independently. The first makes `outOfDate` capable of classifying Identifiers for which only the metadata changed. Building upon that, the second commit refactors `Hakyll.Core.Dependencies` (API-breaking change!) to introduce a `DependencyKind` which allows us to distinguish `KindContent` and `KindMetadata`. The commit also then makes use of the feature where applicable (`getRoute`, `compilerGetMetadata`, `buildTagsWith`). 

Feedback and additional testing would be much appreciated! 🤗

---

Fixes #383

### DIFF
--- a/lib/Hakyll/Core/Dependencies.hs
+++ b/lib/Hakyll/Core/Dependencies.hs
@@ -56,11 +56,12 @@ type DependencyFacts = Map Identifier [Dependency]
 --------------------------------------------------------------------------------
 outOfDate
     :: [Identifier]     -- ^ All known identifiers
-    -> Set Identifier   -- ^ Initially out-of-date resources
+    -> Set Identifier   -- ^ Changed content
+    -> Set Identifier   -- ^ Changed metadata
     -> DependencyFacts  -- ^ Old dependency facts
     -> (Set Identifier, DependencyFacts, [String])
-outOfDate universe ood oldFacts =
-    let (_, state, logs) = runRWS rws universe (DependencyState oldFacts ood)
+outOfDate universe ood oodMeta oldFacts =
+    let (_, state, logs) = runRWS rws universe (DependencyState oldFacts ood oodMeta)
     in (dependencyOod state, dependencyFacts state, logs)
   where
     rws = do
@@ -71,8 +72,9 @@ outOfDate universe ood oldFacts =
 
 --------------------------------------------------------------------------------
 data DependencyState = DependencyState
-    { dependencyFacts :: DependencyFacts
-    , dependencyOod   :: Set Identifier
+    { dependencyFacts   :: DependencyFacts
+    , dependencyOod     :: Set Identifier
+    , dependencyOodMeta :: Set Identifier
     } deriving (Show)
 
 
@@ -90,8 +92,11 @@ markOod id' = State.modify $ \s ->
 -- | Collection of dependencies that should be checked to determine
 -- if an identifier needs rebuilding.
 data Dependencies
-  = DependsOn [Identifier]
+  = DependsOn [(DependencyKind, Identifier)]
   | MustRebuild
+  deriving (Show)
+
+data DependencyKind = KindContent | KindMetadata
   deriving (Show)
 
 instance Semigroup Dependencies where
@@ -108,8 +113,8 @@ dependenciesFor id' = do
     facts <- dependencyFacts <$> State.get
     return $ foldMap dependenciesFor' $ fromMaybe [] $ M.lookup id' facts
   where
-    dependenciesFor' (IdentifierDependency i) = DependsOn [i]
-    dependenciesFor' (PatternDependency _ is) = DependsOn $ S.toList is
+    dependenciesFor' (IdentifierDependency i) = DependsOn [(KindContent, i)]
+    dependenciesFor' (PatternDependency _ is) = DependsOn $ map (\i -> (KindContent, i)) (S.toList is)
     dependenciesFor' AlwaysOutOfDate          = MustRebuild
 
 
@@ -155,12 +160,19 @@ bruteForce = do
         (todo', changed) <- foldM check ([], False) todo
         when changed (go todo')
 
+    findOod oodContent oodMetadata (k, i)
+      = S.member i $
+          case k of
+            KindContent  -> oodContent
+            KindMetadata -> oodMetadata
+
     check (todo, changed) id' = do
         deps <- dependenciesFor id'
         case deps of
           DependsOn depList -> do
-            ood  <- dependencyOod <$> State.get
-            case find (`S.member` ood) depList of
+            ood     <- dependencyOod <$> State.get
+            oodMeta <- dependencyOodMeta <$> State.get
+            case find (findOod ood oodMeta) depList of
                 Nothing -> return (id' : todo, changed)
                 Just d  -> do
                     tell [show id' ++ " is out-of-date because " ++
--- a/lib/Hakyll/Core/Dependencies.hs
+++ b/lib/Hakyll/Core/Dependencies.hs
@@ -1,9 +1,14 @@
 --------------------------------------------------------------------------------
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE TupleSections #-}
 module Hakyll.Core.Dependencies
     ( Dependency (..)
+    , DependencySelector (..)
+    , DependencyKind (..)
     , DependencyFacts
     , outOfDate
+    , contentDependency
+    , metadataDependency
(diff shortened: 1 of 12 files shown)


lib/Hakyll/Core/Dependencies.hs
```diff
@@ -88,6 +106,19 @@ markOod id' = State.modify $ \s ->
...
+
+instance Binary DependencyKind where
+  put KindContent = putWord8 0
+  put KindMetadata = putWord8 1
+
+  get = getWord8 >>= \t -> case t of
+      0 -> pure KindContent
+      1 -> pure KindMetadata
+      _ -> error "Data.Binary.get: Invalid DependencyKind"
+
```
LaurentRDC: How do we ensure that the binary cache is clearly broken by this addition, and does not remain incoherent but still parseable?

One way would be to replace the tokens `0` and `1` by something more explicit, but larger, such as `"kindContent"` and `"kindMetadata"`, but maybe I'm being overly cautious.

lib/Hakyll/Core/Dependencies.hs
```diff
@@ -88,6 +106,19 @@ markOod id' = State.modify $ \s ->
...
+  deriving (Show)
+
+instance Binary DependencyKind where
+  put KindContent = putWord8 0
+  put KindMetadata = putWord8 1
+
+  get = getWord8 >>= \t -> case t of
+      0 -> pure KindContent
+      1 -> pure KindMetadata
+      _ -> error "Data.Binary.get: Invalid DependencyKind"
```
LaurentRDC: Might as well switch to `fail` here as well

lib/Hakyll/Core/Dependencies.hs
```diff
@@ -30,23 +34,37 @@ import           Hakyll.Core.Identifier.Pattern
...
+
+--------------------------------------------------------------------------------
+instance Binary Dependency where
+    put (Dependency k s) = putWord8 0 >> put k >> put s
+    put AlwaysOutOfDate = putWord8 1
+
+    get = getWord8 >>= \t -> case t of
+      0 -> Dependency <$> get <*> get
+      1 -> pure AlwaysOutOfDate
+      _ -> error "Data.Binary.get: Invalid Dependency"
```
LaurentRDC: `error` -> `fail`

LaurentRDC (commented):
Thanks for your contribution! I've got a few comments, no showstoppers.

lib/Hakyll/Core/Dependencies.hs
```diff
@@ -88,6 +106,19 @@ markOod id' = State.modify $ \s ->
...
+data DependencyKind = KindContent | KindMetadata
+  deriving (Show)
+
+instance Binary DependencyKind where
+  put KindContent = putWord8 0
+  put KindMetadata = putWord8 1
+
+  get = getWord8 >>= \t -> case t of
+      0 -> pure KindContent
+      1 -> pure KindMetadata
```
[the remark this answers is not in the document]
nmeum: I wasn't sure how compatibility wrt. the binary cache is handled across versions. I assumed it was recommended to do a clean build after an upgrade anyhow. However, keeping backwards compatibility should be rather trivial if we do something along the lines of:

```Haskell
instance Binary Dependency where
    put (Dependency k s) = putWord8 0 >> put k >> put s
    put AlwaysOutOfDate = putWord8 1

    get = getWord8 >>= \t -> case t of
      -- XXX: Backwards compatability with Hakyll <=4.16.7.1.
      0 -> (\p i -> Dependency KindContent $ PatternDependency p i) <$> get <*> get
      1 -> Dependency KindContent . IdentifierDependency <$> get

      2 -> pure AlwaysOutOfDate
      3 -> Dependency <$> get <*> get
      _ -> fail "Data.Binary.get: Invalid Dependency"
```

Then `0`, `1` (and `3`) will be parsed in the old format and written to disk in the new format. WDYT?

LaurentRDC:
There are more changes than last time I took a look. Was this the result of testing @nmeum ?

nmeum:
No functional changes, I just fixed some code style issues I noticed while reading the PR myself again, added some Haddock comments, rebased and squashed the fixup commits.

LaurentRDC (approved):
Thanks for your contribution!

nmeum:
For those interested in an extended usage example for this feature, I [wrote a tutorial](https://notes.8pit.net/notes/dijr.html) explain how one can generate [a sidebar](https://notes.8pit.net/notes/dijr.html#generating-a-sidebar) and [tag pages](https://notes.8pit.net/notes/dijr.html#generating-tag-pages) using metadata-only dependencies.

UnlimitedCookies:
I may be wrong, but it seems that with this patch something like
```
import Hakyll
[...]
    scssDependencies <- makePatternDependency "css/**.scss"
    rulesExtraDependencies [scssDependencies] $ do
      create ["css/main.scss"] $ do
        route $ setExtension "css"
        compile compressScssCompiler
[...]
```
now looks like this
```
import Hakyll
import Hakyll.Core.Dependencies
[...]
    scssDependencies <- makePatternDependency KindContent "css/**.scss"
    rulesExtraDependencies [scssDependencies] $ do
      create ["css/main.scss"] $ do
        route $ setExtension "css"
        compile compressScssCompiler
[...]
```
So beforehand all that was needed to define a dependency was the prelude, but now the dependency type also needs to be explicitly imported.

Is this intended or am I mistaking?

nmeum:
I guess I just never noticed because all of my source files already had `Hakyll.Core.Dependencies` imported. Could you propose a PR where you re-export `Dependency(..)`, `contentDependency`, and `metadataDependency` from the `Haykll` module?

PR #1085, opened 2025-11-14
getAllMetadata: Depend only on Metadata, not on Content

nmeum:
This refines the dependency tracking of the `getAllMetadata` function using e4b2dfd7d7fe8a6be163638efc8f7f0c9558afbb to only depend on the metadata of matched Items.

For this purpose, it overwrites the default implementation of the `getAllMetadata` function of the `MonadMetadata` type class in the `Compiler` Monad instance.

### DIFF
--- a/lib/Hakyll/Core/Compiler/Internal.hs
+++ b/lib/Hakyll/Core/Compiler/Internal.hs
@@ -39,7 +39,7 @@ module Hakyll.Core.Compiler.Internal
 --------------------------------------------------------------------------------
 import           Control.Applicative            (Alternative (..))
 import           Control.Exception              (SomeException, handle)
-import           Control.Monad                  (forM_)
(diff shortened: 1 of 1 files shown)


lib/Hakyll/Core/Compiler/Internal.hs
```diff
@@ -196,7 +196,12 @@ instance Applicative Compiler where
 -- | Access provided metadata from anywhere
 instance MonadMetadata Compiler where
     getMetadata = compilerGetMetadata
-    getMatches  = compilerGetMatches
+    getMatches  = compilerGetMatches KindContent
+    getAllMetadata pattern = do
```
jaspervdj: ~Name suggestion: `getMatchingMetadata`?~ I thought this PR introduced this function but that's not the case., apologies

LaurentRDC:
Thanks!

PR #1086, opened 2025-11-26
web: add my website as an additional example

nmeum:
I think this might be interesting for others as it is not a blog but rather a personal wiki / public notebook / digital garden and hence requires a somewhat uncommon Hakyll setup.

See: https://notes.8pit.net


LaurentRDC:
Thanks!

Issue #1101, opened 2026-06-01
Hakyll does not plumb through environment variables necessary for locating Typst packages


omentic:
Hello! I am trying to convert my personal wiki over to the Typst format, and am thus adapting my Hakyll build script build script accordingly. Unfortunately, it looks like Hakyll strips some environment variables necessary for Pandoc to compile Typst files that use external packages correctly...

```fish
[wiki]$ ./wiki build
Initialising...
  Creating store...
  Creating provider...
  Running rules...
Checking for out-of-date items
Compiling
  Using async runtime with 2 threads...
  [ERROR] outline.typ: Hakyll.Web.Pandoc.readPandocWith: parse failed: PandocParseError "(line 1, column 2):\nHOME not defined"
make: *** [Makefile:2: build] Error 1
```

Pandoc alone is able to successfully compile the file.
```fish
[wiki]$ pandoc outline.typ -o outline.html
[wiki]$ cat outline.typ 
#import "@preview/cheq:0.3.1": *
#show: checklist

- [ ] linguistics
- [ ] programming language theory
- [ ] mathematics
- [ ] computer science
```

I've never encountered this issue before. Though most formats I've used don't need to actually know the user's HOME directory -- LaTeX packages are installed globally, Markdown doesn't have any packages, etc. Typst does, and I don't think any other major formats really do...

I think Hakyll should plumb around the environment variables to make this succeed successfully by default, so I'm opening an issue. But I think it is also *probably* possible to get this to work within Hakyll's current framework at the moment -- but I don't know how to! The plumbing around of environment variables I've tried to do so far has not been successful... so, in addition to a feature / bug request, I'd also like to request some technical support... 😅

omentic:
Related: #635 (thanks, @Jiramide!)

Following the advice in that issue I was able to define a custom `readPandocWith` [based off the source](https://hackage-content.haskell.org/package/hakyll-4.17.0.0/docs/src/Hakyll.Web.Pandoc.html#readPandocWith), replacing `runPure` with `runIO` and calling `unsafeCompiler` accordingly. I think it's worth considering making `runIO` the default...

```haskell
import Hakyll hiding (pandocCompiler, readPandoc, readPandocWith)

readPandocWith :: ReaderOptions -> Item String -> Compiler (Item Pandoc)
readPandocWith readerOptions item =
  unsafeCompiler $
    runIO (traverse (readWith item readerOptions) (fmap Data.Text.pack item))
    >>= unwrap
  where
[...]

PR #1102, opened 2026-06-04
Run the Pandoc reader in runIO by default

### LINKED ISSUE #635: Hakyll only supports `PandocPure` `PandocMonad`
In `readPandocWith`, `runPure` is used so `pandoc` is run in `PandocPure`. `PandocPure` doesn't support all `pandoc` functionality (in particular, I needed `PandocIO` so that `pandoc` would honor org mode includes).

Options I can imagine: 
- Add functionality to support running `pandoc` in IO (`unsafeCompiler . runIO` seems to Just Work.)
- Document this behavior. It was definitely confusing for me when `hakyll`'s `pandoc` behavior silently differed from directly running `pandoc` (`pandoc` from the CLI defaults to `PandocIO`)
- Examine log messages in [`CommonState`](https://hackage.haskell.org/package/pandoc-2.2/docs/Text-Pandoc-Class.html#t:CommonState) and print them or even crash if they're present

omentic:
This allows for the use of file formats depending on external packages, for example, Typst. 

Closes https://github.com/jaspervdj/hakyll/issues/635, https://github.com/jaspervdj/hakyll/issues/1101

I don't think there's any major downsides to being impure here...?


LaurentRDC:
Don't worry about CI failures and hlint warnings.

I'm still trying to think about potential issues with using `runIO` instead of `runPure` ... but I really can't think of any

jaspervdj:
I think the general reason to avoid `runIO` is that Pandoc / typst / org-mode can include other files in the repository.  This escapes Hakyll's dependency tracking.  So if you have a blogpost `A` which is including e.g. a snippet `B`, `site build` will not rebuild `A` if you make changes to `B`.

If there is a way to have Pandoc produce the files it read as part, we could feed that into the dependency tracker.  Or we could just put a big warning around the fact that includes may not work as expected.

P.S. look forward to meeting you IRL @LaurentRDC!

omentic:
Hmm... how *does* Hakyll's dependency tracking work? Correct me if I'm wrong, but Hakyll's capable of recognizing it needs to rebuild files after local imports like `import path/to/foo.typ` or `image("path/to/image.png")` (or Markdown's `` syntax), right? How does this differ from ex. Org mode includes? 

(And does the use of `runIO` over `runPure` negatively affect this?)

(Conveniently, I think my specific case which I'd like this behaviour for -- the use of external packages in Typst -- sidesteps issues of dependency tracking, as external packages through the `@preview` namespace are explicitly versioned, i.e., if I want to use a different version of an external package, I need to update my file importing it -- which Hakyll will recognize.)

If `runIO` doesn't make any existing cases worse, it's probably worth merging this + documenting the dependency tracking behaviour with remote includes, and then opening another issue for that kind of dependency tracking, I think... though I do wonder if it's possible to pull that kind of information out of Pandoc at all. I didn't see an obvious way from a quick Google search, but I didn't look much into the docs.

omentic:
By the way, @jaspervdj @LaurentRDC what are your thoughts on this now? I still think it would be worthwhile to merge as it's a requirement for using external packages in Typst.

jaspervdj:
I am okay with adding this, but maybe we should add these functions as separate versions (e.g. `readPandocIOWith`).  This will cause some unfortunate duplication, but the alternative is breaking the dependency tracking model that folks could rely on.  Neither are great :-( 

---

I think there are two ways to go about implementing it using dependency tracking:

1.  Use `runPure` without any files.  If anything goes wrong (a file is imported), this will give us a specific `PandocError`.  We can pattern match on this, and add the specific file to the virtual file tree (`insertInFileTree` is pure).  Then retry with this file added.  This could be very slow if you are including thousands of files.
2.  Patch Pandoc to track this and include this info in `CommonState`.  I haven't looked at how hard this would be.  It may be something that is also useful for other Pandoc users though, worth asking.

---

To answer the question:

> it needs to rebuild files after local imports like import path/to/foo.typ or image("path/to/image.png") (or Markdown's  syntax), right? How does this differ from ex. Org mode includes?

I didn't test this, but I suspect Org mode includes don't work either, just like Typst includes.

Markdown's image include syntax works because the output doesn't actually depend on the image contents -- it just writes `` to the output, regardless of what's in the PNG, and regardless of whether or not the PNG actually exists.

omentic:
I'm not sure I completely understand how `runIO` breaks the dependency tracking (nor understand the dependency tracking in general), would you mind elaborating on it a little? Are there any situations in which `runPure` tracks dependencies where `runIO` doesn't? Or is it more that using `runPure` forces you to set up your Hakyll to track dependencies some other way?

---

I would say that option 1 is also bad for the use case I'd most like this for, external Typst packages. Hakyll's dependency tracking doesn't really matter here (and in fact is counterproductive) because packages in the `@preview` namespace are versioned. Unfortunately, the same syntax is used for imports of local files, which are *not* versioned...

I'd suggest an option 3, also, which would be to run specifically Typst in `runIO` and everything else in `runPure`. Most other markup file formats do not have packages and includes in the way it does... though, say, how do LaTeX packages work? Typst installs its packages to `~/.cache/typst/packages/`. If there's a nice way to allow Pandoc to consider those in `runPure`, that might also work (as an option 4).

PR #1104, opened 2026-07-21
Re-export Hakyll.Core.Dependencies from Hakyll module

nmeum:
Otherwise, since #1084, it is not possible to declare dependencies without explicitly importing the `Hakyll.Core.Dependencies` module. Since this was previously possible, it may be worthwhile to also re-export `Hakyll.Core.Dependencies` from the Hakyll module.

See https://github.com/jaspervdj/hakyll/pull/1084#issuecomment-4686676091

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ title: Releases
 ## 4.17.1.0
 
 - Added support for `aeson` 2.3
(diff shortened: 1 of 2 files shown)


LaurentRDC:
That's true, thanks for addressing.

Would you mind updating the changelog as well?

nmeum:
Sorry, forgot about that. Amended the commit accordingly just now.

LaurentRDC:
Thank you!

PR #1106, opened 2026-07-27
Add a new indexRoute for serving extensionless HTML pages

omentic:
Addresses #1039. It's not exactly what I described there, but I think this is a better approach.

This adds a new `Routes` function that serves a `foo.ext` item at `foo/index.html`, allowing its access at `foo`.

It's a very simple piece of code, but I think it would be good to have in Hakyll by default. I don't think it's common knowledge that HTTP servers almost-always support accessing a file at `foo/index.html` by just `foo`.

This should compose fine with other routes via `composeRoutes`.


lib/Hakyll/Core/Routes.hs
```diff
@@ -225,6 +225,15 @@ constRoute :: FilePath -> Routes
 constRoute = customRoute . const
 
 
+--------------------------------------------------------------------------------
+{- | Create a route that routes an item to a index.html file within a folder of
+the name of the item. This is useful for serving files without any extension,
+as most HTTP servers support accessing 'foo/index.html' at 'foo'.
+-}
+indexRoute :: Routes
+indexRoute = customRoute $ (</> "index.html") . dropExtension . toFilePath
```
LaurentRDC, line 234: Can we make this flexible over `"index.html"`? Perhaps with

```haskell
indexRoute :: Routes
indexRoute = indexRouteWith "index.html"

indexRouteWith :: FilePath -> Routes
```

lib/Hakyll/Core/Routes.hs
LaurentRDC, line 1: Is there a way to test this?

lib/Hakyll/Core/Routes.hs
omentic: Well, I don't think there's any benefit to doing so, because anything other than `"index.html"` does not have the property that HTTP servers will serve it for accessing the directory it is contained in, i.e. extensionlessly.

lib/Hakyll/Core/Routes.hs
omentic: I see some tests for `customRoute` that I can replicate. Do you know offhand how to test tests locally?

lib/Hakyll/Core/Routes.hs
LaurentRDC: Actually, now that I think about it, the routing here doesn't make sense. Hakyll isn't a webserver; it generates an output directory, a static website.

Serving `/` as `/index.html` is a property of the server service the static assets. For example, whatever server underlies Github Pages does this automatically. For my personal website, `https://laurentrdc.xyz/` and `https://laurentrdc.xyz/index.html` already render the same page.

What webserver are you using?

lib/Hakyll/Core/Routes.hs
LaurentRDC: You can run all tests using:

```sh
$ cabal test
```

You can run tests whose names match some pattern using:

```sh
$ cabal test --test-options='-p <pattern>'
```

For example:

```sh
$ cabal test --test-options='-p indexRoute'
```

which will run all tests that have the `indexRoute` string infix.

lib/Hakyll/Core/Routes.hs
omentic: Yeah, exactly. This is adding a routing option to make it easy for whatever web server you're using to serve pages extensionlessly, since serving `/index.html` at `/` is a very common property.

The use case I use this for is that I have a bunch of webpages on my site, written in Markdown/Typst and compiled to HTML. With just `setExtension "html"`, webpages like ex. https://toki.la/about -- which are compiled from an `about.md` on disk -- will **only** be accessible at https://toki.la/about.html (with Hakyll's bundled web server).

I don't like this, as I think the `.html` extension is ugly. So this `indexRoute` lets me work around it. I instead pass `indexRoute`, and Hakyll will output `about.html` as `about/index.html` instead. This lets me then access the webpage at https://toki.la/about (without the extension).

Until remembering the common property of HTTP servers to serve files at `about/index.html` as `about/`, I had a lot of UX issues using `hakyll watch` locally, as the upstream web server I'm using on my host *also* will serve `about.html` as just `about` (which is why I didn't bother to look into this more closely before). This is not super common though -- and Hakyll's built-in web server doesn't do it, meaning that all my local links were broken (since they were extensionless URLs).

LaurentRDC: I was confused about what the statically-generated directory structured would look like. Thanks for the explanation.

If I understand correctly, if you want to serve the source file `about.md` as `toki.la/about`, you basically:
* compile `about.md`
* route to `about/index.html`

Ok, that makes sense to me. Can you extend the test suite with this example? Related to my other comment on your PR

omentic: Yup, that's correct. Yeah, I'll extend the test suite after work, thanks for the tips on running them.

mixis:
I have not tested this, but from looking at it, won't this put `index.html` in the `url` metadata field and thus keep it in links on generated archive pages for example? I am currently experimenting with a manual approach where i author pages like `about/index.md`, which should achieve the same result.

Looking for existing solutions, I found [this function for the url field](https://github.com/aherrmann/aherrmann.github.io.hakyll/blob/b7c451d1120dd3b47d4105056bea38b15c1ebd30/src/Main.hs#L189],) also written about in [Jekyll Style URLs with Hakyll](https://aherrmann.github.io/programming/2016/01/31/jekyll-style-urls-with-hakyll/)

I have not considered patching hakyll yet, and don't see how a route sets the url field.

jaspervdj:
This is a great idea, I have written similar helpers many times.  Two remarks:

1. @mixis is right, we would also need a corresponding `Context` to crop out the `/index.html` again if we want generated URLs to look nice.
1.  Consider naming this `indexHtmlRoute`?  I'm not too concerned about having the extension in there, but it makes it more immediately obvious what it's about, especially if we also add the `Context`. `indexContext` could refer to many things, e.g. an sitemap, `indexHtmlContext` is clearer about what it does.  (`appendIndexHtmlRoute` and `trimIndexHtmlContext` could also work but seem to long for my taste.)

